PR #53, opened 2023-07-06
fix: removed url from the WAI sub value, just thumbprint

### LINKED ISSUE #52: Iss in the Wallet Instance Attestation Request should be the same as sub in Response
**iss** in the Wallet Instance Attestation Request payload is defined as:
`The thumbprint of the JWK of the Wallet Instance or which the attestation is  being requested.`

In the payload of the Wallet Instance Attestation **sub** is defined as:
`The public url of the issuer concatenated with the thumbprint of the JWK of the Wallet Instance for which the attestation is being issued.`

Shouldn't they be the same ?

peppelinux:
Roland Hedberg rev

- typo fixed on WAI sub value

### DIFF
--- a/docs/en/wallet-instance-attestation.rst
+++ b/docs/en/wallet-instance-attestation.rst
@@ -388,69 +388,69 @@ Header
 Payload
 ^^^^^^^
 
-+---------------------------+------------------------------------------+
-| **key**                   | **value**                                |
-+---------------------------+------------------------------------------+
-|| iss                      || The public url of the Wallet            |
-||                          || Instance attestation issuer. See        |
-||                          || the example below in this section.      |
-+---------------------------+------------------------------------------+
-|| sub                      || Thumbprint value                        |
-||                          || of the JWK of the Wallet Instance       |
-||                          || for which the attestation is            |
-||                          || being issued.                           |
-+---------------------------+------------------------------------------+
-|| iat                      || Unix timestamp of attestation           |
-||                          || issuance time.                          |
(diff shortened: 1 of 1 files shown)


balanza:
What was the typo? Sorry, it's difficult to spot it as the whole file appears as changed.

balanza:
> What was the typo? Sorry, it's difficult to spot it as the whole file appears as changed.

Ok I got it.

|new|old|
|-|-|
|Thumbprint value of the JWK of the Wallet Instance for which the attestation is being issued.|The public url of the issuer concatenated with the thumbprint of the JWK of the Wallet Instance for which the attestation is being issued.|

It's more a different specification than a typo, isn't it?

peppelinux:
> What was the typo? Sorry, it's difficult to spot it as the whole file appears as changed.

yes, I just ficed the commit since those "automatic" tables are something messy when something changes within them. Now the PR is more readable, thank you for having pointed this out

peppelinux:
> > What was the typo? Sorry, it's difficult to spot it as the whole file appears as changed.
> 
> Ok I got it.
> new 	old
> Thumbprint value of the JWK of the Wallet Instance for which the attestation is being issued. 	The public url of the issuer concatenated with the thumbprint of the JWK of the Wallet Instance for which the attestation is being issued.
> 
> It's more a different specification than a typo, isn't it?

I'd say that's typo due to a regression happened when we have decided to drop the URL from the sub and use just the thumbprint, but feel free to change the title of this PR

grausof (approved):

PR #59, opened 2023-07-08
fix: wallet_provider metadata type

peppelinux:
this PR changes the metadata type name of the wallet provider

`eudi_wallet_provider` -> `wallet_provider`

following the decision already taken and defined in the trust model, here:
https://github.com/italia/eidas-it-wallet-docs/blob/versione-corrente/docs/en/trust.rst#metadata-types


fmarino-ipzs (approved):

grausof (approved):
LGTM

PR #67, opened 2023-07-12
fix!: Wallet Instance Attestation typ value according to VC HAIP

peppelinux:
- WAI typ value changed
- HAIP added to refs
- DPOP refs description updated


grausof (approved):
LGTM

fmarino-ipzs (approved):

PR #77, opened 2023-07-24
feat: [RP] same device flow

### LINKED ISSUE #74: RP Solution - same device flow
In the (Q)EAA Issance flow as defined in the PR #70, we require the Issuer to perform a PID presentation request to the Wallet Instance according to [[OpenID4VP](https://openid.net/specs/openid-4-verifiable-presentations-1_0.html)] following a same-device flow. In the current version of the IT Wallet specification we only have defined a cross-device authorization flow. 

We should add technical details in the Same Device Flow Section.

peppelinux:
This PR closes https://github.com/italia/eudi-wallet-it-docs/issues/74


docs/en/relying-party-solution.rst
```diff
@@ -139,6 +132,22 @@ Below follows its Base64 decoded content:
...
+
+The Relying Party produces a URL similar to the one contained in the QRCode produced in the Cross Device Flow, then provides it to the Wallet Instance via an HTTP 302 response (redirect), represented in the following non-normative example:
+
+.. code:: text
+
+    HTTP/1.1 /pre-authz-endpoint Found
+    Location: https://verifier.example.org/request_uri_endpoint?
+    client_id=https%3A%2F%2Fverifier.example.org%2Fcb
+    &request_uri=https%3A%2F%2Fverifier.example.org%2Frequest_uri_endpoint
+
```
grausof: Shall we add that the user will first have to select their wallet provider?

peppelinux: Why?

in the presentation phase the user interact directly to the RP with their wallet instance. 

What would you like to say?

grausof: if the user has multiple wallet providers (multiple verifier apps) how do they decide which one to use?
The universal link is unique for each verifier app

fmarino-ipzs: Maybe we should better clarify the scenario. I see at least the following scenarios:

1. The User starts a browser (external to the Walle Instance) in the device, wants to get access to a RP and the RP trigger the Wallet Instance for the User authentication...
2. The User interacts directly with the Wallet Instance that provide a list of RP services to the User. The User select the service to which requesting the access and the Wallet Instance starts an embedded browser for the interaction with the RP.

In the case 1 we need the User select the Wallet Provider directly into the RP. In the second case, the wallet instance could handle the redirect as the User is already in the Wallet Instance.

peppelinux: Please, we really have to take out the possibility that the user should select the wallet provider in a RP discovery page.

The RP provides the QRCode, the User open their Wallet Instance and starts the flow.
At the end of the road we must adopt all the compromises to have the EUDI interop, then every member state will support the same things (even with custom extension but the same for interop needs)

for the PoC the user knows that a specific app is needed to start the flow

fmarino-ipzs: Ok, we should (or MUST??) avoid burdening the user with this. At the same time we should understand how to handle the Wallet Instance invocation in an external initiated scenario. I think this is the point. We could reach a compromise for the PoC but I think we should have a clear proposal on this topic regarding the EUDI interoperability.

docs/en/relying-party-solution.rst
```diff
@@ -102,6 +98,18 @@ The payload of the QR Code is a **Base64 encoded string** based on the following
 
   eudiw://authorize?client_id=`$client_id`&request_uri=`$request_uri`
 
+In the Same Device Flow the URL is provided as described in the previous section, where the parameter **client_id** and **request_uri** are the same if the ones used in the Cross Device Flow and the only difference if in the url schema and in the removal of the Verifier's FQDN
+
```
```suggestion
In the Same Device Flow the parameter **client_id** and **request_uri** are the same if the ones used in the Cross Device Flow with the only difference about the url schema and the removal of the Verifier's FQDN from the URL. 

```

docs/en/relying-party-solution.rst
```diff
@@ -102,6 +98,18 @@ The payload of the QR Code is a **Base64 encoded string** based on the following
...
+This Same Device Flow utilizes HTTP redirects to finalize the authorization phase and obtain Verifiable Presentation(s).
```
```suggestion
In the Same Device Flow the Relying Party uses a HTTP response redirect (status code 302) to give to the Wallet Instance the resource where the request object is available for download, as represented in the following non-normative example:
```

docs/en/relying-party-solution.rst
```diff
@@ -102,6 +98,18 @@ The payload of the QR Code is a **Base64 encoded string** based on the following
...
+
+The Relying Party produces the URL to be provided to the Wallet Instance using a HTTP 302 response (redirect), as represented in the following non-normative example:
```
```suggestion
```

docs/en/relying-party-solution.rst
```diff
@@ -58,11 +53,11 @@ Once the Relying Party authentication is performed by the Wallet Instance, the U
   * - **2**
     - The Relying Party creates an Authorization Request which contains the scopes of the request.
   * - **3**
-    - The Relying Party inserts the reference URI of the *request_uri* into a QR Code.
+    - In the **Same Device Flow** the Relying Party responses with the Request URI in the form of HTTP Location (302). In the **Cross Device Flow** the Request URI is provided in the form of a QR Code.
```
grausof, line 56: This is valid if there is only one verifier. If there is more than one then RP will have to make the user choose his Wallet Provider to provide the correct redirect url.

grausof (approved):

fmarino-ipzs (approved):

docs/en/relying-party-solution.rst
peppelinux: @grausof 

I'm going to merge this, your comment was moved here:
https://github.com/italia/eudi-wallet-it-docs/issues/81

PR #85, opened 2023-07-30
fix!: credential_definition.type is a single value

peppelinux:
There's a small deviation from the sd-jwt vc spec, given by the authorization details example that uses the parameter type as an array and not as literal string.

when we have defined it we assumed that the type should be multivalued, this have been consolidated now in the sd-jwt vc using the single valued approach.

See also [here](https://github.com/vcstuff/oid4vc-haip-sd-jwt-vc/pull/56/files#diff-15671165287eaf4ce50e0013632c722709931937f92d73118d2265657914f61f).

### DIFF
--- a/docs/en/pid-eaa-issuance.rst
+++ b/docs/en/pid-eaa-issuance.rst
@@ -137,7 +137,7 @@ The JWS payload of the request object is represented below:
         "type":"openid_credential",
         "format": "vc+sd-jwt",
         "credential_definition": {
-            "type": ["PersonIdentificationData"]
+            "type": "PersonIdentificationData"
         }
     }
     ],


docs/en/pid-eaa-issuance.rst
```diff
@@ -137,7 +137,7 @@ The JWS payload of the request object is represented below:
         "type":"openid_credential",
         "format": "vc+sd-jwt",
         "credential_definition": {
-            "type": ["PersonIdentificationData"]
+            "type": "PersonIdentificationData"
         }
     }
     ],
```
peppelinux, line 143: @fmarino-ipzs

`authorization_details` is an array, and each credential authorization is a json object, then requesting multiple credentials is possible by specifing type, format and credential definition.

fair enough I'd say

balanza (commented):
> a small deviation from the sd-jwt vc spec

Given this deviation, should we briefly explain this choice with a small paragraph in the docs? That may be helpful for readers to reckon we are still in the realm of sd-jwt, and we took a motivated choice.

---

> See also [here](https://github.com/vcstuff/oid4vc-haip-sd-jwt-vc/pull/56/files#diff-15671165287eaf4ce50e0013632c722709931937f92d73118d2265657914f61f).

That PR is still open and the discussion is still ongoing. Should we wait before assuming it right?

peppelinux:
@balanza "that deviation" is resolved in this PR

it was that singlevalued vs the multivalued, as it was before

balanza (approved):

ruphy (approved):
LGTM

grausof (approved):
LGTM

fmarino-ipzs (approved):

PR #89, opened 2023-08-02
Relying Party security considerations for Cross Device Flow

peppelinux:
This PR uses the same url scheme used in the cross device also for the same device

it also adds a section related to the security considerations about how a cross device flow should secure the user-agent session


docs/en/relying-party-solution.rst
```diff
@@ -98,15 +98,15 @@ The payload of the QR Code is a **Base64 encoded string** based on the following
...
+In the Same Device Flow the parameter **client_id** and **request_uri** are the same if the ones used in the Cross Device Flow with the only difference about the url schema and the removal of the Verifier's FQDN from the URL.
 
 In the Same Device Flow the Relying Party uses a HTTP response redirect (status code 302) to give to the Wallet Instance the resource where the request object is available for download, as represented in the following non-normative example:
 
 
 .. code:: text
 
     HTTP/1.1 /pre-authz-endpoint Found
-    Location: https://verifier.example.org/request_uri_endpoint?
+    Location: eudiw://authorize?
```
grausof: Please leave https:// which is a Universal link for the app. You cannot use a deep link. We'll make a proposal on that but I wouldn't change the schema for now

peppelinux: ok, so a change is then required since we cannot keep verifier.example.org, isn't that true?

do you have any proposal for this universal link, like `https://eudi.wallet.gov.it`?

grausof: Correct, our idea was to have a universal link (managed by trust anchors) that brings together all the wallet providers (an RFC on this will follow). So for now an idea could be` https://wallet-providers.eudi.wallet.gov.it`

docs/en/relying-party-solution.rst
```diff
@@ -140,14 +140,36 @@ Below is a non-normative example of the QR Code raw payload:
...
 .. code-block:: text
 
-  ZXVkaXc6Ly9hdXRob3JpemU/Y2xpZW50X2lkPWh0dHBzOi8vdmVyaWZpZXIuZXhhbXBsZS5vcmcmcmVxdWVzdF91cmk9aHR0cHM6Ly92ZXJpZmllci5leGFtcGxlLm9yZy9yZXF1ZXN0X3VyaQ==
+  ZXVkaXc6Ly9hdXRob3JpemU/Y2xpZW50X2lkPWh0dHBzOi8vdmVyaWZpZXIuZXhhbXBsZS5vcmcmcmVxdWVzdF91cmk9aHR0cHM6Ly92ZXJpZmllci5leGFtcGxlLm9yZy9yZXF1ZXN0X3VyaS8=
 
 Below follows its Base64 decoded content:
 
 .. code-block:: text
 
   eudiw://authorize?client_id=https://verifier.example.org&request_uri=https://verifier.example.org/request_uri
```
grausof: I understand that this will be the content of the QR code. If so, how will the RP backend distinguish one request from another?
In this way the qr code will be static and identical for everyone

grausof (changes requested):

docs/en/relying-party-solution.rst
balanza: Why don't we just leave it unchanged for now? So we can merge this PR.

We can then propose changes in a new PR

@peppelinux

docs/en/relying-party-solution.rst
peppelinux: it is a non normative example but anyway we should decide it the reuqest_uri is the same but with urlparams ?state=$that-state or having dynamic request_uri (and then outside of the wallet relying party metadata) that are related to different request_objects (available in the queue)

I think that it would be better taking the JARM recommendation, remove the request_uri from the metadata, and have random request_uri bound to specific state.

@grausof do you agree?

grausof (approved):
LGTM

fmarino-ipzs (approved):

PR #91, opened 2023-08-04
Relying Party sequence diagram, examples and editorials

peppelinux:
this PR made several improvements in the sequence diagrams and editorials, adding more details and reorganizes the contents.


docs/en/relying-party-solution.rst
```diff
@@ -16,139 +16,124 @@ This section describes how a Relying Party may ask to a Wallet Instance the pres
...
 
 Remote Protocol Flow
 --------------------
-In the **Same Device** and **Cross Device** Authorization Flows, the User interacts with a remote Relying Party.
-This scenario requests the Verifier to provide the URL where the signed request object is available.
 
-The Verifier MUST detect the device type of the requestor (Wallet Instance), if it is a mobile device or a workstationm and activate one the supported remote flows:
+In this scenario the Relying Party provides the URL where the signed presentation request object is available for download.
+
+The Relying Party MUST detect the device type of the requestor (Wallet Instance), if it is a mobile device or a workstation, and activate one the supported remote flows:
```
grausof: Put this way it looks like the Wallet Instance could be a workstation as well.

```suggestion
Depending on whether the Relying Party client is on a mobile device or a workstation, it must activate one of the supported remote flows:
```

grausof (changes requested):

docs/en/relying-party-solution.rst
```diff
@@ -16,139 +16,123 @@ This section describes how a Relying Party may ask to a Wallet Instance the pres
 
 In this section the following flows are described:
 
-- **Remote Same Device Flow**, where the Verifier and the Wallet Instance acts in the same device.
-- **Remote Cross Device Flow**, where the Verifier and the Wallet Instance acts in different devices and the Verifier is a remote Relying Party.
+- **Remote Same Device Flow**, where the Web Browser and the Wallet Instance are used in the same device.
```
grausof: Why Web Browser? Couldn't it also be an app? I propose a generic name

peppelinux: I suppose that even if embedded in the app it is always a web browser (better: user agent)

may we say just user-agent?

grausof: better!

balanza: In some contexts it's used the term `Reader Instance` as the _abstract_ agent that performs a single interaction with the Relying Party (that is: a Reader Instance relates to Relying Party as Wallet Intance relates to Wallet Provider).

peppelinux: Reader Instance is defined for proximity flows while in this section we only cover remote flows

```suggestion
- **Remote Same Device Flow**, where the user-agent and the Wallet Instance are used in the same device.
```

docs/en/relying-party-solution.rst
```diff
@@ -16,139 +16,123 @@ This section describes how a Relying Party may ask to a Wallet Instance the pres
...
+- **Remote Cross Device Flow**, where the Web Browser and the Wallet Instance are used in different devices.
```
```suggestion
- **Remote Cross Device Flow**, where the user-agent and the Wallet Instance are used in different devices.
```

docs/en/relying-party-solution.rst
balanza: I see. It's just a proposal to use a name that does not bring any technological constraint/expectation. User-agent is fine; in same-device flow, we can have web applications (browser) or native applications.

balanza (approved):

ruphy (approved):

PR #92, opened 2023-08-05
fix: trust model, rp ES example, removed AS reserved parameters

peppelinux:
This PR removes the AS reserved parameters that was included in the example of openid relying party Entity Statement, within the metadata policies

Thanks to Takahiko Kawasaki for this!


balanza (approved):

grausof (approved):
LGTM

PR #98, opened 2023-08-25
fix: relying party status endpoint made simpler

peppelinux:
This PR makes the section related to the status endpoint more readable, it also aligns the proposed http status codes to the current implementation of pyeudiw

### DIFF
--- a/docs/en/relying-party-solution.rst
+++ b/docs/en/relying-party-solution.rst
@@ -129,7 +129,7 @@ The decoded content of the previous Base64 value is represented below:
 
 .. code-block:: text
 
-  eudiw://authorize?client_id=https://relying-party.example.org&request_uri=https%3A%2F%2Frelying-party.example.org%2Frequest_uri%3Fid%3Drandom-value
+  eudiw://authorize?client_id=https%3A%2F%2Frelying-party.example.org&request_uri=https%3A%2F%2Frelying-party.example.org%2Frequest_uri%3Fid%3Drandom-value
 
 .. note::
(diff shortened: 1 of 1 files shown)


grausof (approved):

docs/en/relying-party-solution.rst
```diff
@@ -142,16 +142,14 @@ When the flow is Cross Device, the user-agent needs to check the session status
 
 Since the QRcode page and the status endpoint are implemented by the Relying Party, it is under its responsability the implementation details of this solution, since it is related to the Relying Party's internal API.
 
-The Relyng Party MUST bind the request of the user-agent, with a Secured and Httponly session cookie, with the issued request (using the ``random-value`` parameter within the ``request_uri`` value). The HTTP response returned by this specialized endpoint MAY contain the HTTP status codes listed below:
+The Relyng Party MUST bind the request of the user-agent, with a Secured and Httponly session cookie, with the issued request. The request url SHOULD include a parameter with a random value. The HTTP response returned by this specialized endpoint MAY contain the HTTP status codes listed below:
```
balanza, line 145: **question:** ‏_SHOULD_ means it can be omitted. What happens id we omit the parameter?

balanza (approved):

docs/en/relying-party-solution.rst
peppelinux: we not mandate an implementation to have a random value since the lookup parameter can be implemented on top of a cookie, we mandate the security requirements but now how these must be implemented to be satisfied. However we define a proposal with a normative language that leaves up to the implementation the technical solution

fmarino-ipzs (approved):

PR #100, opened 2023-08-29
fix: openid credential issuer discovery example

### LINKED ISSUE #96: Federation discovery of PID/(Q)EAA issuers
from Roland

In https://italia.github.io/eudi-wallet-it-docs/versione-corrente/en/pid-eaa-issuance.html
HIgh-level PID flow.

The document says: "the Wallet Instance queries the Trust Anchor to fetch the trusted PID Provider."
I guess it should say "... the identifier for the trusted PID Provider."

------------------------

I agree that the discovery process should be described better, assuming also that there may be multiple (Q)EAA issuers registered with TA's Intermediates

peppelinux:
Resolves https://github.com/italia/eudi-wallet-it-docs/issues/96


docs/en/pid-eaa-issuance.rst
```diff
@@ -37,7 +37,7 @@ The :numref:`fig_High-Level-Flow-ITWallet-PID-Issuance` shows a general architec
 Below a detailed description for each step represented in the previous picture:
 
     0. **Wallet Instance Setup**: the first time the Wallet Instance is started a preliminary setup phase MUST be carried out. It consists of the release of a verifiable proof issued by the Attestation Service provided by the Wallet Provider that asserts the genuineness, the authenticity and the compliance with a trust framework of the Wallet Instance. The verifiable proof binds a public key corresponding to a local private key generated by the Wallet Instance. 
-    1. **Obtaining the trusted PID Provider**: the Wallet Instance queries the Trust Anchor to fetch the trusted PID Provider. 
+    1. **Obtaining the trusted PID Provider**: the Wallet Instance discovers the trusted PID Provider, eg: `https://trust-anchor.eudi.wallet.developers.italia.it/list?entity_type=openid_credential_issuer` and then inspects the metadata looking to the availability of the PID credential.
```
peppelinux: Following @rohe suggestions

Obtaining the trusted PID Provider has these steps:

- Use the federation list interface with a query like what you specify: `https://trust-anchor.eudi.wallet.developers.italia.it/list?entity_type=openid_credential_issuer`

- Then for each of the entity_ids returned collect their metadata and look for what ? credentials_supported:credential_definition:type == ??

@fmarino-ipzs I think that we should specify in the second point the domestic PID namespace used for discovering the credential availability upon a specific issuer

fmarino-ipzs: In the PID Provider Metadata we have `credential_supported:credential_definition:type=PersonIdentificationData`.

fmarino-ipzs: Anyway, I think that it is better to add a new section describing the discovery phase, with some non-normative examples, for the sake of readability. What do you think? This is a high-level view I wouldn't go into details here.

fmarino-ipzs:
```suggestion
    1. **Obtaining the trusted PID Provider**: the Wallet Instance discovers the trusted PID Provider at the Subordinate Listing Endpoint of the Trust Anchor, and then inspects the metadata looking for the availability of the PID credential. 
```

peppelinux: Yes, I agree

fmarino-ipzs (approved):

PR #113, opened 2023-09-05
fix!: removing unused and deprecated WIA claims

peppelinux:
According to the chat with Bundes friends, we agreed to lowring the density of the WIA, also considering the updates in 

https://github.com/vcstuff/draft-ietf-oauth-attestation-based-client-auth/blob/main/draft-ietf-oauth-attestation-based-client-auth.md#wallet-instance-attestation


balanza (approved):

grausof (approved):
LGTM

PR #120, opened 2023-09-07
Wallet Instance Attestation Request header content type

### LINKED ISSUE #105: Wallet Instance request media type
@rohe 's

`var+jwt` (Verifiable Assertion Request JWT) are there other Verifiable Assertion Request then a Wallet Instance Attestation Request?
If so I'd rather have `wiar+jwt` as the header `typ` for a Wallet Instance Attestation Request.

We should be as specific as necessary.

peppelinux:
this PR sets the JWS header typ and removes the claim `type` in the payload


grausof (approved):
LGTM

fmarino-ipzs (approved):

PR #157, opened 2023-11-02
fix: mistakes and ambiguities removed related to PE and scopes

peppelinux:
Re-reading the RP section I found several mistakes about the usage of scope and PE

I removed them and I would merge this as it is and at the same time continuing working on the following

- https://github.com/openid/OpenID4VCI/issues/89
- https://github.com/italia/eudi-wallet-it-docs/issues/152

### DIFF
--- a/docs/en/relying-party-solution.rst
+++ b/docs/en/relying-party-solution.rst
@@ -284,7 +284,7 @@ The Relying Party issues the signed Request Object, where a non-normative exampl
   }
   .
   {
-    "scope": "eu.europa.ec.eudiw.pid.it.1 email",
+    "scope": "eu.europa.ec.eudiw.pid.it.1 tax_id_number",
     "client_id_scheme": "entity_id",
     "client_id": "https://relying-party.example.org",
     "response_mode": "direct_post.jwt",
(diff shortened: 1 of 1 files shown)


docs/en/relying-party-solution.rst
```diff
@@ -288,7 +284,7 @@ The Relying Party issues the signed Request Object, where a non-normative exampl
   }
   .
   {
-    "scope": "eu.europa.ec.eudiw.pid.it.1 pid-sd-jwt:unique_id+given_name+family_name",
+    "scope": "eu.europa.ec.eudiw.pid.it.1 email",
```
Sh-Amir: A very minor point: by the time the only optional parameter that we introduced in the PID Data Model is the user Tax  code ``tax_id_number``. Maybe it is better to use that otherwise, we have an inconsistency with what we reported in that section.

Sh-Amir (commented):
change ``scope``: ``eu.europa.ec.eudiw.pid.it.1 email`` to ``eu.europa.ec.eudiw.pid.it.1 tax_id_number``

docs/en/relying-party-solution.rst
```suggestion
    "scope": "eu.europa.ec.eudiw.pid.it.1 tax_id_number",
```

peppelinux: Done, thx

fmarino-ipzs (approved):

grausof (approved):

PR #170, opened 2023-11-23
fix: credential_definition type -> vct

peppelinux:
According to https://vcstuff.github.io/oid4vc-haip-sd-jwt-vc/draft-oid4vc-haip-sd-jwt-vc.html#name-authorization-details

@fmarino-ipzs I think that this PR should be extended with some words in the text that says something about this requirement

if W3C dt is used, `type` would be then used. I'd create a warning box ...

### DIFF
--- a/docs/en/pid-eaa-issuance.rst
+++ b/docs/en/pid-eaa-issuance.rst
@@ -177,7 +177,7 @@ The JWS of Request Object is represented below:
       "type":"openid_credential",
       "format": "vc+sd-jwt",
       "credential_definition": {
-          "type": "PersonIdentificationData"
+          "vct": "PersonIdentificationData"
       }
   }
   ],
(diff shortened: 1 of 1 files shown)


fmarino-ipzs:
I see your point, but this is not aligned with https://openid.github.io/OpenID4VCI/openid-4-verifiable-credential-issuance-wg-draft.html#name-credential-issuer-metadata-2. 
Am I missing something?

peppelinux:
@fmarino-ipzs `type` is for W3C dt 1.1, while `vct` is for sd-jwt-vc ...

peppelinux:
@fmarino-ipzs profile merged here
https://github.com/openid/OpenID4VCI/pull/128/files

peppelinux:
Closed by https://github.com/openid/OpenID4VCI/commit/12d5ede25056daba991906e845d80128ad57572c

PR #182, opened 2023-12-28
fix: Wallet Instance Attestation renamed to Wallet Attestation

peppelinux:
This PR renames Wallet Instance Attestation to Wallet Attestation since OpenID specs uses Wallet Attestation and eIDAS has abandoned the term WIA for Wallet Trust Evidence.

We're prone to use the term defined in the openid specs.

This change was discussed during the workshop in Rome.


fmarino-ipzs (approved):

docs/en/wallet-instance-attestation.rst
grausof, line 1: what do you think about renaming the file too?

grausof (approved):
LGTM

docs/en/wallet-instance-attestation.rst
peppelinux: ok

peppelinux: done https://github.com/italia/eudi-wallet-it-docs/pull/182/commits/65c86e944d73761dce27bba82d1570af6bdba3f1

aarmam:
> eIDAS has abandoned the term WIA for Wallet Trust Evidence.

@peppelinux
In ARF 1.4 I still see separate definitions for Wallet Instance Attestation (used for wallet instance revocation) and Wallet Trust Evidence (used to prove to Issuer that the key to bind with Credential can be trusted and the private key is protected by a certified
WSCA/WSCD).

peppelinux:
@aarmam indeed. My comment was related to the conversation had 5 months ago

first the name was wallet instance attestation, then it was changed to wallet attestation, then the WTE was introduced and the wallet attestation appeared a sort of it, then the term wallet attestation was moved again to wallet instance attestation

your comment is correct.

PR #207, opened 2024-02-11
fix: rp presentation definition typo

peppelinux:
This PR fixes the non-normative example of the RP presentation definition


fmarino-ipzs (approved):

balanza (approved):
LGMT

PR #221, opened 2024-03-01
Status Attestation Responses

balanza:
Refactor the descriptions of the HTTP responses for easier reads.

Key changes:
* all the possible HTTP status codes for a response are listed in the table so that readers can have a complete reference at a glimpse
* avoid response body when no useful information is returned to the client (example: in 5xx errors)
* error codes are listed in a dedicated table

Solves https://github.com/italia/eudi-wallet-it-docs/issues/219

### DIFF
--- a/docs/en/revocation-lists.rst
+++ b/docs/en/revocation-lists.rst
@@ -200,10 +200,52 @@ The Revocation Endpoint MUST be provided by the Issuer within its Metadata.
 Credential Revocation HTTP Response
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-The response MUST be an HTTP Response using the status code set to *204 (No-Content)* in the case of successful revocation. If the Digital Credential could not be found by the Issuer, an HTTP Response with status code *404 (Not Found)* MUST be returned. Otherwise an HTTP error response MUST be provided by the Issuer to the Wallet Instance. This response MUST use *application/json* as the content type and MUST include the following parameters:
+In case of successful revocation, the Issuer MUST return an HTTP Response with status code *204 No Content*.
+Otherwise, an HTTP error response MUST be provided by the Issuer using status codes according to the table below. 
 
-  - *error*. The error code.
+To revoke an already revoked Digital Credential will result in a successful response.
+
+The following HTTP Status Codes MUST be supported:
+
+.. list-table:: 
+    :widths: 20 20 60
+    :header-rows: 1
+
+    * - **Status Code**
+      - **Body**
+      - **Description**
+    * - *204 No Content*
+      - 
+      - The Digital Credential has been successfully revoked.
+    * - *400 Bad Request*
+      - Error code and description
+      - The issuer cannot fulfill the request because of invalid parameters.
+    * - *404 Not Found*
+      - 
+      - The Digital Credential to be revoked can not be found by the Issuer.
+    * - *500 Internal Server Error*
+      - 
+      - The Issuer encountered an internal problem. (:rfc:`6749#section-5.2`).
+    * - *503 Service Unavailable*
+      - 
+      - The Issuer is temporary unavailable. (:rfc:`6749#section-5.2`).
+
+For HTTP error responses that involve a body, the body MUST be encoded in ``application/json`` format and MUST contain the following parameters:
+
+  - *error*. The error code, as registerd in the table below.
   - *error_description*. Text in human-readable form providing further details to clarify the nature of the error encountered.
+
+The following Error Codes MUST be supported:
+
+.. list-table:: 
+    :widths: 20 80
+    :header-rows: 1
+
+    * - **Error Code**
+      - **Description**
+    * - ``invalid_request``
+      - The request is not valid due to the lack or incorrectness of one or more parameters. (:rfc:`6749#section-5.2`).
+
 
 Below a non-normative example of an HTTP Response with an error.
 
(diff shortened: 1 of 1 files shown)


balanza:
Please note I tried to be more clear on the use of error codes:
> Error codes are meant to provide additional information about the failure so that the User can be informed and take the appropriate action.

I wanted to enforce the concept that the error code MUST NOT revert the meaning of the http response (which is a failure).

docs/en/revocation-lists.rst
```diff
@@ -200,42 +200,65 @@ The Revocation Endpoint MUST be provided by the Issuer within its Metadata.
...
-      - **Error Code**
+      - **Body**
       - **Description**
+    * - *204 No Content*
+      - 
+      - The Digital Credential has been successfully revoked.
     * - *400 Bad Request*
-      - *invalid_request*
-      - The request is not valid due to the lack or incorrectness of one or more parameters. (:rfc:`6749#section-5.2`).
+      - Error code and description
```
peppelinux, line 219: is this intended to be here?

if yes, why it is not in the similar errors below?

docs/en/revocation-lists.rst
```diff
@@ -359,48 +382,62 @@ The *Credential status endpoint* MUST return a response with a *HTTP status code
...
+    * - ``invalid_request``
+      - The request is not valid due to the lack or incorrectness of one or more parameters. (:rfc:`6749#section-5.2`).
+    * - ``credential_revoked``
       - The Digital Credential is revoked. The reason of revocation MUST be provided in the *error_description* field.
-    * - *400 Bad Request*
-      - *credential_updated*
+    * - ``credential_updated``
       - One or more attributes contained in the Digital Credential are changed. The *error_description* field MUST contain a list of updated attributes.
 
+Below a non-normative example of an HTTP Response with an error.
```
peppelinux, line 430: duplicated: https://github.com/italia/eudi-wallet-it-docs/pull/221/files#diff-d3bfad4aa613161f55b96d232893081a0ac30d5127aca04dda54925b43aceb1eR250

peppelinux (changes requested):

docs/en/revocation-lists.rst
balanza: I don't understand this comment.

balanza: Ok, maybe I understood what you meant. 

The response body is required only for the cases in which it effectively adds information for the user. That is, there's no value in sending "server error" as a payload to a 500 Server Error response.

Also, 5xx errors are not always handled by the application but rather by gateways and proxies, so it may be impractical to customize them. As it brings no value, I'd rather not consider the body.

docs/en/revocation-lists.rst
balanza: They are the same snippets but they apply to different endpoints.

I think it's ok to have examples when you expect them, even if they are duplicated

docs/en/revocation-lists.rst
fmarino-ipzs: I agree with @balanza , the only status code that makes sense to include a body in is the _400 Bad Request_ status code. Having said that, probably we could remove the body column saying that only the 400 (Bad Request) status Code MUST include a body encoded in ``application/json`` format and MUST contain the following parameters: [...].

@peppelinux  WDYT?

grausof (approved):
LGTM

docs/en/revocation-lists.rst
peppelinux: no, since we have server_error in RFC 6749
https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.2.1

docs/en/revocation-lists.rst
peppelinux: I can approve and merge as it is 

in the future we'll try to make some editorials to make the text shorter and use more references

peppelinux (approved):

docs/en/revocation-lists.rst
```diff
@@ -359,48 +381,63 @@ The *Credential status endpoint* MUST return a response with a *HTTP status code
...
-      - *invalid_request*
-      - The request is not valid due to the lack or incorrectness of one or more parameters. (:rfc:`6749#section-5.2`).
+      - Error code and description
+      - The issuer cannot fulfill the request because of invalid parameters.
+    * - *404 Not Found*
+      - 
+      - The Digital Credential can not be found by the Issuer.
     * - *500 Internal Server Error*
-      - *server_error*
+      -
```
peppelinux, line 403: this is RFC6749

https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.2.1

peppelinux (changes requested):

docs/en/revocation-lists.rst
```diff
@@ -359,48 +381,63 @@ The *Credential status endpoint* MUST return a response with a *HTTP status code
       - It MUST contain the Status Attestation as a signed JWT. 
       - `[OAuth Status Attestation draft 01] <https://datatracker.ietf.org/doc/draft-demarco-status-attestations/01/>`_.
 
-If the Digital Credential could not be found by the Issuer, an HTTP Response with status code 404 (Not Found) MUST be returned. In all other cases the Issuer MUST return an HTTP Response Error using *application/json* as the content type, and including the following parameters:
```
peppelinux, line 362: why it was removed?

peppelinux (changes requested):

docs/en/revocation-lists.rst
fmarino-ipzs: Yes but here probably we can omit it as in the RFC6749 I read:

_(This error code is needed because a 500 Internal Server Error HTTP status code cannot be returned to the client via an HTTP redirect.)_

Am I missing something?
Anyway, if we don't reach a consensus now, I suggest to put this PR to the next release if you agree.

docs/en/revocation-lists.rst
fmarino-ipzs: It is simply moved below, @balanza is it right?

peppelinux (approved):

docs/en/revocation-lists.rst
balanza: I can add the error payload if we feel it's worth it, it doesn't affect the way the communication works.

Let me just point out that:
* RFC6749 does not design the endpoint to use multiple http codes as error, that's why it needs to differentiate by payload
* In RFC6749, the http response's content type is `application/x-www-form-urlencoded`, we are using `application/json`
* 5xx will usually be generated by systems other than the resource server (ex: load balancers, api gateways, firewalls, etc), so it's probable we will have payload not in the expected format.

docs/en/revocation-lists.rst
balanza: Details on every possible response have been moved in the response table: https://github.com/italia/eudi-wallet-it-docs/pull/221/files/fdd6d6fa93100f45a71fbab1dc082a5f85c176fa#diff-d3bfad4aa613161f55b96d232893081a0ac30d5127aca04dda54925b43aceb1eR401

Rationale: it gives the same information in a structured way, using fewer words. Implementers may like it.

fmarino-ipzs (approved):

PR #227, opened 2024-03-02
Trust infrastructure functional requirements and general properties

peppelinux:
This PR extends the section of the trust infrastructure, looking to the future integrations with X.509 and trusted lists

it also reduce and update the introduction of this specs


docs/en/trust.rst
```diff
@@ -165,7 +204,7 @@ All the endpoints listed below are defined in the `OIDC-FED`_ specs.
 | historical keys           | **GET**                                      |                                |  Trust Anchor   |
 |                           |                                              |Lists the expired and revoked   |                 |
 |                           |                                              |keys, with the motivation of the|  Intermediate   |
-|                           | .well-known/openid-federation-historical-jwks|revocation.                     |                 |
+|                           | /historical-jwks                             |revocation.                     |                 |
```
balanza, line 207: **nitpick:** ‏If there's no specific reason I would keep the `.well-known` prefix as it's consistent with the metadata endpoint.

```suggestion
|                           | .well-known/historical-jwks                             |revocation.                     |                 |
```

balanza (approved):
LGTM

docs/en/trust.rst
peppelinux: unfortunately it is something defined (changed) in the openid federation specs.

peppelinux: but I got your point in terms of normalization. unfortunately the endpoint must be defined in the federation_entity metadata and the correct URL must be taken from there, despite its normalization as well-known

fmarino-ipzs (approved):

PR #267, opened 2024-04-16
chore: add recommedation about public key used in the jwt proof

peppelinux:
This PR aims to remark an important aspect about the unlinkability of the credential through the holder key binding.

### DIFF
--- a/docs/en/pid-eaa-issuance.rst
+++ b/docs/en/pid-eaa-issuance.rst
@@ -324,7 +324,7 @@ The ``client_assertion`` is signed using the private key that is created during 
 
 .. note::
 
-    **It is RECOMMENDED that the public key contained in the ``jwt_proof`` be specifically generated for the requested Credential (fresh cryptographic key) to ensure that different issued Credentials do not share the same public key, thereby remaining unlinkable to each other.
+    It is RECOMMENDED that the public key contained in the ``jwt_proof`` be specifically generated for the requested Credential (fresh cryptographic key) to ensure that different issued Credentials do not share the same public key, thereby remaining unlinkable to each other.
 
 .. code-block::
 
--- a/docs/en/pid-eaa-issuance.rst
+++ b/docs/en/pid-eaa-issuance.rst
@@ -322,6 +322,7 @@ The ``client_assertion`` is signed using the private key that is created during 
 
     **PID/(Q)EAA Credential Schema and Status registration:** The PID/(Q)EAA Provider MUST register all the issued Credentials for their later revocation, if needed.
 
+
 .. note::
 
     It is RECOMMENDED that the public key contained in the ``jwt_proof`` be specifically generated for the requested Credential (fresh cryptographic key) to ensure that different issued Credentials do not share the same public key, thereby remaining unlinkable to each other.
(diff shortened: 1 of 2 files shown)


docs/en/pid-eaa-issuance.rst
```diff
@@ -322,6 +322,10 @@ The ``client_assertion`` is signed using the private key that is created during
 
     **PID/(Q)EAA Credential Schema and Status registration:** The PID/(Q)EAA Provider MUST register all the issued Credentials for their later revocation, if needed.
 
+.. note::
+
+    **It is RECOMMENDED that the public key contained in the ``jwt_proof`` be specifically generated for the requested Credential (fresh cryptographic key) to ensure that different issued Credentials do not share the same public key, thereby remaining unlinkable to each other.
```
peppelinux:
```suggestion
    It is RECOMMENDED that the public key contained in the ``jwt_proof`` be specifically generated for the requested Credential (fresh cryptographic key) to ensure that different issued Credentials do not share the same public key, thereby remaining unlinkable to each other.
```

docs/en/pid-eaa-issuance.rst
```diff
@@ -322,6 +322,10 @@ The ``client_assertion`` is signed using the private key that is created during
 
     **PID/(Q)EAA Credential Schema and Status registration:** The PID/(Q)EAA Provider MUST register all the issued Credentials for their later revocation, if needed.
```
peppelinux, line 324:
```suggestion


```

aarmam:
I will add the same option for discussion, to solve enforcing the key to bind for CI, as described in openid/OpenID4VCI#292

> OpenID4VCI has an option to use authorization_details parameter defined by [Section 2 of RFC9396](https://www.rfc-editor.org/rfc/rfc9396.html#name-request-parameter-authoriza). The same specification could be used to convey [authorization_details in token endpoint](https://www.rfc-editor.org/rfc/rfc9396.html#name-token-request) to specify the authorization details that a client wants the AS to assign to an access token. For example credential_jkt parameter could be introduced to force CI to accept proofs that match credential_jkt claim in access token.

This would enforce the CI (required by OpenID4VCI if they choose to implement it or by EUDI Wallet implementation profile otherwise) to verify that the key thumbprint in credential_jkt matches the key thumbprint in proof parameter.

> The name and intention is similar to [RFC 9449 dpop_jkt parameter](https://datatracker.ietf.org/doc/html/rfc9449.html#dpop_jkt), that you use also for sender-constrained access tokens, that can be used to bind the issued authorization code to a specific key. 

I dont know if its overkill but you could consider using dpop_jkt in PAR request similary to enforce the DPoP key that can be presented in token endpoint for sender-constrained access token. But that is a separate issue.

OpenID4VCI states that
> The Client can request issuance of a Credential of a certain type multiple times, e.g., to associate the Credential with different public keys/Decentralized Identifiers (DIDs) or to refresh a certain Credential

This could be achieved with authorization_details in a refresh token request to specify updated value of credential_jkt.

fmarino-ipzs (approved):

grausof (approved):

PR #277, opened 2024-05-13
fix: sd-jwt vc sub value

### LINKED ISSUE #276: sub parameter in SD-JWT-VC
According to [draft-ietf-oauth-sd-jwt-vc-02](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-sd-jwt-vc-02#name-registered-jwt-claims) the claim `sub` is 

_The identifier of the Subject of the Verifiable Credential. The Issuer MAY use it to provide the Subject identifier known by the Issuer._

We should fix it in the section [PID/(Q)EAA SD-JWT parameters](https://italia.github.io/eudi-wallet-it-docs/wallet-attestation/en/pid-eaa-data-model.html)

peppelinux:
This PR aims to resolve the issue https://github.com/italia/eudi-wallet-it-docs/issues/276

the solution proposed goes along the experience made with openid connect, where the sub value identifying the user is opaque and not reversible.

### DIFF
--- a/docs/en/pid-eaa-data-model.rst
+++ b/docs/en/pid-eaa-data-model.rst
@@ -85,7 +85,7 @@ The following claims MUST be in the JWT payload. Some of these claims can be dis
       - [NSD].URL string representing the PID/(Q)EAA Issuer unique identifier.
       - `[RFC7519, Section 4.1.1] <https://www.iana.org/go/rfc7519>`_.
     * - **sub**
-      - [NSD].Thumbprint of the JWK in the ``cnf`` parameter.
+      - [NSD].The identifier of the subject of the Digital Credential, the User, must be opaque and should not correspond to any anagraphic data or be derived from the User's anagraphic data through pseudonymization.
       - `[RFC7519, Section 4.1.2] <https://www.iana.org/go/rfc7519>`_.
     * - **iat**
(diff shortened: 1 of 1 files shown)


grausof (approved):

docs/en/pid-eaa-data-model.rst
```diff
@@ -85,7 +85,7 @@ The following claims MUST be in the JWT payload. Some of these claims can be dis
       - [NSD].URL string representing the PID/(Q)EAA Issuer unique identifier.
       - `[RFC7519, Section 4.1.1] <https://www.iana.org/go/rfc7519>`_.
     * - **sub**
-      - [NSD].Thumbprint of the JWK in the ``cnf`` parameter.
+      - [NSD].The identifier of the subject of the Digital Credential, the User, must be opaque and should not correspond to any anagraphic data or be derived from the User's anagraphic data through pseudonymization.
```
peppelinux:
```suggestion
      - [NSD]. The identifier of the subject of the Digital Credential, the User, MUST be opaque and MUST NOT correspond to any anagraphic data or be derived from the User's anagraphic data via pseudonymization. Additionally, it is required that two different Credentials issued MUST NOT use the same ``sub`` value.
```

fmarino-ipzs (approved):

Issue #257, opened 2024-04-15
[Wallet Attestation - Requirements ] Req.9 Loss of Wallet Istance private key


pietroACN:
The requirement states: "When the private key associated with the Wallet Instance is lost or deleted, the attestation MUST become invalid to prevent unauthorized use of the Wallet Instance".  [Link ](https://github.com/italia/eudi-wallet-it-docs/blob/df948503e22cbf949530a07b6cee0e83e6901613/docs/en/wallet-attestation.rst?plain=1#L33C1-L33C168)

This requirement is unclear:
1) Wallet Attestation is signed by Wallet Provider, while this requirement refers to Wallet Instance private key.
2) If this key lost how does the wallet owner can act within its wallet and 'invalidate' this attestation? 
3) How does the attestation 'becomes' invalid? 
4) If owner communicates asynchronously this information to Wallet Provider (for example via a specific properly authenticated web page managed by WP itself) user will not be able to identify himself via his private key as it's lost, thus will need similar (L3) methods to securely identify. In any case how WP will be able to send an update to the wallet about Wallet Attestation revocation?

This requirement needs to be well detailed with a proper asynchronous process to avoid misuse.

peppelinux:
1. WIA is signed by WP, at the same time the WIA binds the WI's public key (cnf.jwk). This means that at least two keys are involved for this issuance.
2. several channels must be taken into account, since there's the need to allow users that lost their phones to be able to deactivate them. This is out of the scopes of the specification, while represent security impacts within the wallet solution. Therefore a better explainataion about, at east, this requirement and some implementations consideration would help
3. it can be revoked or it is simply short-lived. At this stage in the italian specs we only deal with short liveld attestations
4. legacy eID systems and/or remote id proofing methods must be involved in these cases. The revocation is related to the wallet instance and not to its wallet attestation. This can be further analyzied in terms of security impacts.

please provide the solution you consider valuable for the resolution of this issue

PR #300, opened 2024-05-31
Wallet Instance Revocation

cmarco0:
this PR aims to solve #257


docs/en/wallet-attestation.rst
```diff
@@ -559,7 +559,20 @@ The details of the revocation mechanism used by the Wallet Provider as well as t
...
 
+- The smartphone is lost;
+- The smartphone has been compromised (e.g., a malicious actor gains control of the smartphone);
+- The smartphone has been reset to factory settings;
+- Other scenarios where the owner can no longer access their Wallet Instance.
+
+If any of these situations occur, the owner **MUST** revoke the Wallet Instance.
+To achieve this the Wallet Provider (WP) **MUST** provide a Web Portal where the end user can access and request the revocation of the Wallet Instance.
+
+The suggested solution for a owner/end user is to establish a secure backup plan at the beginning of the registration process. The backup **MUST NOT** be stored in the same location as the private key. Instead, it should be saved to a different location, such as a cloud solution or a legacy storage solution. The purpose of the backup is to facilitate the recovery of the Hardware Private Key generated by the secure HSM chip on the smartphone's motherboard. For example, a method such as a the generation of seed phrases could be implemented to recover the lost Hardware private key.
```
```suggestion
```

docs/en/wallet-attestation.rst
```diff
@@ -559,7 +559,20 @@ The details of the revocation mechanism used by the Wallet Provider as well as t
...
+In this manner, the owner can effortlessly restore their Wallet Instance without generating a new one.
```
```suggestion
```

docs/en/wallet-attestation.rst
```diff
@@ -559,7 +559,20 @@ The details of the revocation mechanism used by the Wallet Provider as well as t
...
+
+The backup and restore specification are address in the *Backup and Restore* chapter.
```
```suggestion
```

docs/en/wallet-attestation.rst
```diff
@@ -559,7 +559,20 @@ The details of the revocation mechanism used by the Wallet Provider as well as t
...
 When the User consents to being linked with the Wallet Instance, they gain the ability to directly request Wallet revocation from the Wallet Provider, and it also allows the Wallet Provider to revoke the Wallet Instance associated with that User.
 
+The following situations could occur:
 
+- The smartphone is lost;
+- The smartphone has been compromised (e.g., a malicious actor gains control of the smartphone);
+- The smartphone has been reset to factory settings;
+- Other scenarios where the owner can no longer access their Wallet Instance.
+
+If any of these situations occur, the owner **MUST** revoke the Wallet Instance.
```
```suggestion
If any of these situations occur, the owner **MUST** revoke the Wallet Instance.
```

please define the "owner"

pietroACN: I would suggest to use 'holder' in every occurrence of 'owner' .
Holder would be defined as: "User that has activated and has full control of an instance of the EUDI wallet and the related private keys".
Some reference of this can be seen in art. 6a, point 7 of the amending regulation

docs/en/wallet-attestation.rst
pietroACN: I would remove the part of the sentence, e.g removing " The backup **MUST NOT** be stored in the same location as the private key. Instead, it should be saved to a different location, such as a cloud solution or a legacy storage solution."
From one side, backup could be within the same WSCD device for example if this is a local external device, and a generic 'cloud solution' surely is not advisable at this purpose as it could not provide the right level of security.

docs/en/wallet-attestation.rst
```diff
@@ -559,7 +559,25 @@ The details of the revocation mechanism used by the Wallet Provider as well as t
 During the *Wallet Instance initialization and registration* phase the Wallet Provider MAY associate the Wallet Instance with a specific User, subject to obtaining the User's consent. The Wallet Provider MUST evaluate the operating system and general technical capabilities of the device to check compliance with the technical and security requirements and to produce the Wallet Instance metadata.
 When the User consents to being linked with the Wallet Instance, they gain the ability to directly request Wallet revocation from the Wallet Provider, and it also allows the Wallet Provider to revoke the Wallet Instance associated with that User.
 
+The following situations could occur:
```
```suggestion
Regarding the reasons for revoking a Wallet Instance, the following scenarios may occur:
```

docs/en/wallet-attestation.rst
```diff
@@ -559,7 +559,25 @@ The details of the revocation mechanism used by the Wallet Provider as well as t
...
 
+- The smartphone is lost;
+- The smartphone has been compromised (e.g., a malicious actor gains control of the smartphone);
+- The smartphone has been reset to factory settings;
+- Other scenarios where the holder can no longer access their Wallet Instance.
```
```suggestion
- Any other scenarios where the User loses the control of the Wallet Instance.
```

docs/en/wallet-attestation.rst
```diff
@@ -559,7 +559,25 @@ The details of the revocation mechanism used by the Wallet Provider as well as t
...
+
+If any of these situations occur, the holder **MUST** revoke the Wallet Instance.
```
```suggestion
If any of the previous scenarios occur, the Wallet Instance **MUST** be revoked.
```

docs/en/wallet-attestation.rst
```diff
@@ -559,7 +559,25 @@ The details of the revocation mechanism used by the Wallet Provider as well as t
...
+
+.. note::
+
+the *Holder* is defined as the user that has activated and has full control of an instance of the EUDI wallet and the related private keys
```
```suggestion
```

we have the section with the defined terms for this. Please do not define terms along the documents sections

docs/en/wallet-attestation.rst
```diff
@@ -559,7 +559,25 @@ The details of the revocation mechanism used by the Wallet Provider as well as t
...
+The following situations could occur:
 
+- The smartphone is lost;
+- The smartphone has been compromised (e.g., a malicious actor gains control of the smartphone);
+- The smartphone has been reset to factory settings;
+- Other scenarios where the holder can no longer access their Wallet Instance.
+
+If any of these situations occur, the holder **MUST** revoke the Wallet Instance.
+
+.. note::
```
```suggestion
```

we don't use normative language within the notes, since the notes must be used as notes only.

docs/en/wallet-attestation.rst
```diff
@@ -559,7 +559,25 @@ The details of the revocation mechanism used by the Wallet Provider as well as t
...
+
+the *Holder* is defined as the user that has activated and has full control of an instance of the EUDI wallet and the related private keys
+
+To achieve this the Wallet Provider (WP) **MUST** provide a Web Portal where the end user can access and request the revocation of the Wallet Instance.
```
```suggestion
To allow the User to revoke the Wallet Instance, the Wallet Provider (WP) **MUST**  offer a remote service, such as a web page, where the User can authenticate and request the revocation of a previously activated Wallet Instance.
```

docs/en/wallet-attestation.rst
```diff
@@ -559,7 +559,23 @@ The details of the revocation mechanism used by the Wallet Provider as well as t
...
 
+- The smartphone is lost;
+- The smartphone has been compromised (e.g., a malicious actor gains control of the smartphone);
+- The smartphone has been reset to factory settings;
+- Any other scenarios where the User loses the control of the Wallet Instance.
+
+If any of the previous scenarios occur, the Wallet Instance **MUST** be revoked.
+
+
+
```
```suggestion
```

docs/en/wallet-attestation.rst
```diff
@@ -559,7 +559,23 @@ The details of the revocation mechanism used by the Wallet Provider as well as t
...
+To allow the User to revoke the Wallet Instance, the Wallet Provider (WP) **MUST**  offer a remote service, such as a web page, where the User can authenticate and request the revocation of a previously activated Wallet Instance.
+
+The suggested solution for a holder is to establish a secure backup plan at the beginning of the registration process. The purpose of the backup is to facilitate the recovery of the Hardware Private Key generated by the secure HSM chip on the smartphone's motherboard. For example, a method such as a the generation of seed phrases could be implemented to recover the lost Hardware private key.
```
```suggestion
```

docs/en/wallet-attestation.rst
```diff
@@ -559,7 +559,23 @@ The details of the revocation mechanism used by the Wallet Provider as well as t
...
+In this manner, the holder can effortlessly restore their Wallet Instance without generating a new one.
+
```
```suggestion
```

out of scope for the section wallet attestation.

docs/en/wallet-attestation.rst
peppelinux: we already had a conversation about this and it makes me frustrated found it again and be forced to remove it by my hands.

the scope of the technical specification is not provide "suggested solution" but define what must or must not, should or should not or may be done with clear statements. key derivation mechanisms are not standardized and therefore out of this specs. At the same time, in the form proposed in this PR are not useful.

grausof (approved):

fmarino-ipzs (approved):

PR #307, opened 2024-06-04
Update credential revocation reason from issuer's perspective

### LINKED ISSUE #266: [Credential Lifecycle - Status Attestation ] - Functional Requirements - Issuer
The text states: "  The issuer Must .... revoke a Digital Credential when ....the Wallet Instance that holds the Digital Credential was issued is revoked;"

It is not clear where is the relation between the  Wallet Istance (that may have been revoked and replaced with another version) and the Digital Credential as the credential validity should be independent from the wallet: the credential is bound to the user Key that may be generated by an external WCSD or remote signing device.

It would be preferable to change the text with the following one: "  The issuer Must .... revoke a Digital Credential when ....the cryptographic key material to which the issued Digital Credential is bound is lost or replaced;"

In case also the following [statement ](https://github.com/italia/eudi-wallet-it-docs/blob/df948503e22cbf949530a07b6cee0e83e6901613/docs/en/revocation-lists.rst?plain=1#L90C1-L90C37)should be replaced with: "Loss of cryptographic key material to which the issued Digital Credential is bound"

cmarco0:
this PR aims to solve this issue https://github.com/italia/eudi-wallet-it-docs/issues/266

### DIFF
--- a/docs/en/revocation-lists.rst
+++ b/docs/en/revocation-lists.rst
@@ -57,7 +57,7 @@ Functional Requirements
 - revoke a Digital Credential when the following circumstances occur:
 
   - the Digital Credential requires to be updated, whenever one or more attributes are changed; in this case the User will request a new issuance for that Digital Credential;
-  - the Holder needs to address the loss or compromise of cryptographic key material associated with the issued Digital Credential. In such case, the Holder should request revocation through a web service provided by the Wallet Provider, using an authentication method that offers a High Level of Assurance (e.g., CIE ID level 3);
+  - the Holder needs to address the loss or compromise of cryptographic key material associated with the issued Digital Credential. In such case, the End-User should request the revocation of the Digital Credential through a service provided by the Credential Issuer and using an authentication method that offers the same Level of Assurance obtained during the Credential Issuance;
   - the User deletes the Digital Credential from the Wallet Instance;
 
 - provide a web service for allowing a Wallet Instance, with a proof of possession of a specific Digital Credential, to 
--- a/docs/en/revocation-lists.rst
+++ b/docs/en/revocation-lists.rst
@@ -58,7 +58,7 @@ Functional Requirements
 
   - the Digital Credential requires to be updated, whenever one or more attributes are changed; in this case the User will request a new issuance for that Digital Credential;
   - the Holder needs to address the loss or compromise of cryptographic key material associated with the issued Digital Credential. In such case, the End-User should request the revocation of the Digital Credential through a service provided by the Credential Issuer and using an authentication method that offers the same Level of Assurance obtained during the Credential Issuance;
-  - the User deletes the Digital Credential from the Wallet Instance;
+  - the User deletes the Digital Credential from the Wallet Instance. The Wallet Instance therefore should request the revocation of such Digital Credential to the Credential Issuer;
 
 - provide a web service for allowing a Wallet Instance, with a proof of possession of a specific Digital Credential, to 
 
--- a/docs/en/revocation-lists.rst
+++ b/docs/en/revocation-lists.rst
@@ -57,7 +57,7 @@ Functional Requirements
 - revoke a Digital Credential when the following circumstances occur:
 
   - the Digital Credential requires to be updated, whenever one or more attributes are changed; in this case the User will request a new issuance for that Digital Credential;
-  - the Holder requests such revocation due to the loss or replacement of cryptographic key material to which the issued Digital Credential is bound. In such case, the only way a Holder can request a revocation is to use an alternative authentication method with a High Level of Assurance (e.g., CIE level 3);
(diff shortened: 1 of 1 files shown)


docs/en/revocation-lists.rst
```diff
@@ -57,7 +57,7 @@ Functional Requirements
 - revoke a Digital Credential when the following circumstances occur:
 
   - the Digital Credential requires to be updated, whenever one or more attributes are changed; in this case the User will request a new issuance for that Digital Credential;
-  - the Wallet Instance that holds the Digital Credential was issued is revoked;
+  - the holder request such revocation due to loss or replacement of cryptographic key material which the issued Digital Credential is bound to;
```
peppelinux:
```suggestion
  - the Holder request such revocation due to loss or replacement of cryptographic key material which the issued Digital Credential is bound to;
```

how it can request the revocation if it cannot prove to be the legitimate possessor of that credential?

fmarino-ipzs: The only way could be making the request directly to the Issuer which must authenticate the User.

cmarco0: Something with a high level of assurance, like CIE (maybe only by using CIE level 3)

peppelinux: shall we say this in the text?

docs/en/revocation-lists.rst
```diff
@@ -57,7 +57,7 @@ Functional Requirements
 - revoke a Digital Credential when the following circumstances occur:
 
   - the Digital Credential requires to be updated, whenever one or more attributes are changed; in this case the User will request a new issuance for that Digital Credential;
-  - the holder request such revocation due to loss or replacement of cryptographic key material which the issued Digital Credential is bound to;
+  - the Holder requests such revocation due to the loss or replacement of cryptographic key material to which the issued Digital Credential is bound. In such case, the only way a Holder can request a revocation is to use an alternative authentication method with a High Level of Assurance (e.g., CIE level 3);
```
peppelinux:
```suggestion
  - the Holder requests such revocation due to the loss or replacement of cryptographic key material to which the issued Digital Credential is bound. In such case, the Holder should request a revocation using a web service provisioned by the Wallet Provider and an authentication method with a High Level of Assurance (e.g., CIE id level 3);
```

peppelinux (approved):

docs/en/revocation-lists.rst
```diff
@@ -57,7 +57,7 @@ Functional Requirements
 - revoke a Digital Credential when the following circumstances occur:
 
   - the Digital Credential requires to be updated, whenever one or more attributes are changed; in this case the User will request a new issuance for that Digital Credential;
-  - the Wallet Instance that holds the Digital Credential was issued is revoked;
+  - the Holder requests such revocation due to the loss or replacement of cryptographic key material to which the issued Digital Credential is bound. In such case, the Holder should request a revocation using a web service provisioned by the Wallet Provider and an authentication method with a High Level of Assurance (e.g., CIE id level 3);
```
peppelinux: `replacement of cryptographic key material to which the issued Digital Credential is bound` sounds misleading, if the digital credential has a binded cryptographic material this is not supposed to be replaced

peppelinux:
```suggestion
  - the Holder needs to address the loss or compromise of cryptographic key material associated with the issued Digital Credential. In such case, the Holder should request revocation through a web service provided by the Wallet Provider, using an authentication method that offers a High Level of Assurance (e.g., CIE ID level 3);
```

docs/en/revocation-lists.rst
```diff
@@ -57,7 +57,7 @@ Functional Requirements
 - revoke a Digital Credential when the following circumstances occur:
 
   - the Digital Credential requires to be updated, whenever one or more attributes are changed; in this case the User will request a new issuance for that Digital Credential;
-  - the Wallet Instance that holds the Digital Credential was issued is revoked;
+  - the Holder needs to address the loss or compromise of cryptographic key material associated with the issued Digital Credential. In such case, the Holder should request revocation through a web service provided by the Wallet Provider, using an authentication method that offers a High Level of Assurance (e.g., CIE ID level 3);
```
grausof:
```suggestion
  - the Holder needs to address the loss or compromise of cryptographic key material associated with the issued Digital Credential. In such case, the End-User should request the revocation of the Digital Credential through a service provided by the Credential Issuer and using an authentication method that offers the same Level of Assurance obtained during the Credential Issuance;
```

docs/en/revocation-lists.rst
```diff
@@ -57,7 +57,7 @@ Functional Requirements
...
   - the User deletes the Digital Credential from the Wallet Instance;
```
peppelinux:
```suggestion
  - the User deletes the Digital Credential from the Wallet Instance. The Wallet Instance therefore should request the revocation of such Digital Credential to the Credential Issuer;
```

grausof (approved):

fmarino-ipzs (approved):

PR #315, opened 2024-06-13
Created a presentation_definition JSON validator python script

cmarco0:
This script is designed to validate JSON data against a specified JSON schema and additional custom rules. It reads the JSON schema and data from files, checks if the data conforms to the schema, and ensures that certain additional requirements are met. This validation process helps in ensuring that the JSON data structure adheres to expected standards and rules, making it reliable and consistent for further processing or usage in applications.


script/JSON_validator/presentation_definition_validator.py
```diff
@@ -0,0 +1,94 @@
+import json
```
peppelinux: this is not what I have explicitly asked

we need a loader of the json schema validator using pydantic, as the provided example. We don't need static code doing static validation since the json validation schema may change

script/JSON_validator/err1.png
peppelinux, line 1: we don't need images

peppelinux (changes requested):

peppelinux:
presentation definition is not deprecated and not part anymore of openid4vp

PR #364, opened 2024-07-12
[Relying Party Solution - Remote Flow - Authorization Response Details]

### LINKED ISSUE #359: [Relying Party Solution - Remote Flow - Authorization Response Details]
In this process the Wallet Instance sends the Authorization Response to the Relying Party.
It would be necessary to clarify the "response " format changing the example set [here ](https://github.com/italia/eudi-wallet-it-docs/blob/1cdf40bb3e4cac2627acb2668c797d3e0a6ae759/docs/en/remote-flow.rst?plain=1#L445) from "Below is a non-normative example of the decrypted JSON response content" to "Below is a non-normative example of the decrypted payload of the JWT, before base64url encoding and signing"

SaraConsoliACN:
This PR aims to resolve the issue https://github.com/italia/eudi-wallet-it-docs/issues/359

### DIFF
--- a/docs/en/remote-flow.rst
+++ b/docs/en/remote-flow.rst
@@ -397,8 +397,7 @@ Below a non-normative example of the request:
   response=eyJhbGciOiJFUzI1NiIs...9t2LQ
   
 
-Below is a non-normative example of the decrypted JSON ``response`` content:
-
+Below is a non-normative example of the decrypted payload of the JWT contained in the ``response``, before base64url encoding and signing:
 .. code-block:: 
 
   {
--- a/docs/en/remote-flow.rst
+++ b/docs/en/remote-flow.rst
@@ -397,7 +397,7 @@ Below a non-normative example of the request:
   response=eyJhbGciOiJFUzI1NiIs...9t2LQ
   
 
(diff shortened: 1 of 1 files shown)


peppelinux (approved):

docs/en/remote-flow.rst
```diff
@@ -397,7 +397,7 @@ Below a non-normative example of the request:
   response=eyJhbGciOiJFUzI1NiIs...9t2LQ
   
 
-Below is a non-normative example of the decrypted JSON ``response`` content:
+Below is a non-normative example of the decrypted payload of the JWT contained in the ``response``, before base64url encoding and signing:
```
peppelinux:
```suggestion
Below is a non-normative example of the decrypted payload of the JWT contained in the ``response``, before base64url encoding:
```

it is not clear how the RP may obtain the wallet instance public keys if it will not ask the WIA

fmarino-ipzs: In our current approach the RP may ask for the Wallet Attestation in the scope parameter within the request object. We may change approach if you want but I suggest to open an issue for this.

fmarino-ipzs (approved):
LGTM after the suggestion commit of @peppelinux

grausof (approved):

PR #384, opened 2024-08-06
fix: Add `sub` claim to Wallet Attestation 

### LINKED ISSUE #360: Wallet Attestation registered claim does not include sub
In the (non normative) example Wallet Attestation, which can be found here
https://github.com/italia/eudi-wallet-it-docs/blob/e6bc7e37d8573d310c08d30efe112be0d50f15b8/docs/en/wallet-attestation.rst?plain=1#L324-L371

the claim `sub` appears with value equal to the jwk thumbprint of the key to be confirmed. On the other hand, the claim `sub` is not listed in the Walklet Attestation registered claim table, which can be found here
https://github.com/italia/eudi-wallet-it-docs/blob/e6bc7e37d8573d310c08d30efe112be0d50f15b8/docs/en/wallet-attestation.rst?plain=1#L474-L515

It not clear to me wether the `sub` claim should appear or not, and what value should it have when it appears.
If I understand correctly, this is a [partial reference](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-attestation-based-client-auth-02#name-client-attestation-jwt), referenced by [this other reference](https://openid.github.io/oid4vc-haip-sd-jwt-vc/openid4vc-high-assurance-interoperability-profile-sd-jwt-vc-wg-draft.html#name-wallet-attestation-schema) suggest that `sub` claim should appear. If it must appear, it is still not clear to me wether its value should be the jwk thumprint of the `cnf` key or not.

grausof:
Resolves https://github.com/italia/eudi-wallet-it-docs/issues/360

### DIFF
--- a/docs/en/wallet-attestation.rst
+++ b/docs/en/wallet-attestation.rst
@@ -486,6 +486,9 @@ The body of the Wallet Attestation JWT MUST contain:
     * - **iss**
       - Identifier of the Wallet Provider
       - :rfc:`9126` and :rfc:`7519`.
+    * - **sub**
+      - Identifier of the Wallet Instance which is the thumbprint of the Wallet Instance JWK.
+      - :rfc:`9126` and :rfc:`7519`.
     * - **exp**
       - UNIX Timestamp with the expiry time of the JWT.
       - :rfc:`9126` and :rfc:`7519`.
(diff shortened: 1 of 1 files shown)


docs/en/wallet-attestation.rst
```diff
@@ -486,6 +486,9 @@ The body of the Wallet Attestation JWT MUST contain:
     * - **iss**
       - Identifier of the Wallet Provider
       - :rfc:`9126` and :rfc:`7519`.
+    * - **sub**
+      - Identifier of the Wallet Instance which is the thumbprint of the Wallet Instance JWK.
```
```suggestion
      - Identifier of the Wallet Instance which is the thumbprint of the Wallet Instance JWK contained in the ``cnf`` claim.
```

fmarino-ipzs (approved):

peppelinux (approved):

PR #412, opened 2024-09-15
fix: kid is not required in sd-jwt hkb

### LINKED ISSUE #411: Redundant parameter in header of Key Binding of SD-JWT
In the header of the Key Binding JWT the parameter `kid` is included
https://github.com/italia/eudi-wallet-it-docs/blob/491419ca1fc537fa04eda0fccb5106393e565ca6/docs/en/remote-flow.rst?plain=1#L477-L478
but the parameter is NOT actually included in the reference specs of [sd-jwt#section5.3](https://www.ietf.org/archive/id/draft-ietf-oauth-selective-disclosure-jwt-10.html#name-key-binding-jwt)

This make sense since the signature verification key is [always identified without ambiguity](https://www.rfc-editor.org/rfc/rfc7800.html#section-3.1) in the `cnf` claim of the issuer jwt, hence no `kid` need to be matched. Similarly, `kid` is also usually (but not necessarily) omitted from the `cnf.jwk` claim of the issuer jwt - this also holds in the it specs, see
https://github.com/italia/eudi-wallet-it-docs/blob/491419ca1fc537fa04eda0fccb5106393e565ca6/examples/pid-sd-jwt-example-payload.json#L29-L36

peppelinux:
This PR resolves https://github.com/italia/eudi-wallet-it-docs/issues/411

### DIFF
--- a/docs/en/remote-flow.rst
+++ b/docs/en/remote-flow.rst
@@ -474,8 +474,6 @@ When an SD-JWT is presented, its KB-JWT MUST contain the following parameters in
     - REQUIRED. MUST be ``kb+jwt``, which explicitly types the Key Binding JWT as recommended in Section 3.11 of [RFC8725].
   * - **alg**
     - REQUIRED. Signature Algorithm using one of the specified in the section Cryptographic Algorithms. 
-  * - **kid**
-    - REQUIRED. Unique identifier of the public key to be used to verify the signature.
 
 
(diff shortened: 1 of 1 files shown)


fmarino-ipzs (approved):
I also removed the `kid` param in the text

m-basili (approved):

grausof (approved):

PR #416, opened 2024-09-18
breaking-change: universal links are not used

### LINKED ISSUE #383: [Relying Party Solution] [Remote Flow] Unclear definition of authorize request
In the non-normative example of RP auhorization request, to the best of my knowledge there are up to 3 issues.
https://github.com/italia/eudi-wallet-it-docs/blob/6d06ccfd551d005a2211c4489936fc2faa0c7fa8/docs/en/remote-flow.rst?plain=1#L163
(1) the scheme invocation is `https://`, but accordin to [HAIP](https://openid.net/specs/openid4vc-high-assurance-interoperability-profile-sd-jwt-vc-1_0.html#name-openid-for-verifiable-prese:~:text=Section%207.2.-,As%20a%20way%20to%20invoke%20the%20Wallet%2C%20at%20least%20a%20custom%20URL%20scheme%20haip%3A//%20MUST%20be%20supported,-.%20Implementations%20MAY%20support) `haip://` should be preferred when the Wallet Instance is a mobile app.
(2) It is not clear what the host of the request should be. In the example, it is `wallet-solution.digital-strategy.europa.eu`. I'll refrain from making a sugegstion as this is beyond my area of expertise.
(3) the authorization endpoint of the example request is `/authorization` but the default authorization endpoint in OAuth is `/authorize`. Unless there is a particular reason to change the authorization endpoint, I would stick the example with the default.

peppelinux:
This PR fixes https://github.com/italia/eudi-wallet-it-docs/issues/383

it has moves part of the original content provised in https://github.com/italia/eudi-wallet-it-docs/pull/400

This PR needs to be extended with support of universal linsk as well, as requested by @grausof 

This PR needs to support both openid4vp:// and haip://


grausof:
Outdated, there is already an issue linked to this https://github.com/italia/eudi-wallet-it-docs/issues/240

PR #430, opened 2024-10-01
chore: Add missing params on Attestation Request

### LINKED ISSUE #365: Claims carried from the attestation request to the attestation
According to the spec the body of the Wallet Attestation JWT MUST contain among other claims:

- authorization_endpoint
- response_types_supported
- response_modes_supported
- vp_formats_supported
- request_object_signing_alg_values_supported
- presentation_definition_uri_supported

The only way the wallet provider can have gotten this information is by the attestation request and there the claims above are
not listed as required.
A slight mismatch. 
Either the above listed claims should be required in both the request and the assertion or they should be optional.

grausof:
Fix https://github.com/italia/eudi-wallet-it-docs/issues/365

### DIFF
--- a/docs/en/wallet-attestation.rst
+++ b/docs/en/wallet-attestation.rst
@@ -246,7 +246,7 @@ Below a non-normative example of the ``client_data``.
 
 **Steps 11-12**: The Wallet Instance:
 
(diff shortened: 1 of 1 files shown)


docs/en/wallet-attestation.rst
```diff
@@ -283,6 +283,17 @@ Below an non-normative example of the Wallet Attestation Request JWT without enc
         },
       },
     },
+    authorization_endpoint": "eudiw:",
```
```suggestion
    authorization_endpoint": "https://wallet-solution.digital-strategy.europa.eu/authorization",
```

peppelinux (approved):

m-basili (approved):

PR #460, opened 2024-10-25
typ in req obj jwt

### LINKED ISSUE #458: Unecessary generic wording of Request Object jwt header `typ`
The request object JWT header must include a `typ` header as specificied here
https://github.com/italia/eudi-wallet-it-docs/blob/fdd9cf44e69bd6eeb4cfb11684dd6f5e0e7c6a57/docs/en/remote-flow.rst?plain=1#L277-L278
but the spec does not state what value `typ` must have, relying instead on a generic description.

It is implicit from the non normative example that `typ` must be `JWT`, but IMO if this is the case then the description should state it.

Note that, surprisingly, [RFC9101 (JAR)](https://www.rfc-editor.org/rfc/rfc9101.html) does not actually mandate the usage of `typ` and indeed it is NOT included in the non-normative example. This is because a Request Object is always delivered in a HTTP response with media type `application/oauth-authz-req+jwt`, making the `typ` field not required.

Zicchio:
This PR resolves #458 by stating that `typ` MUST be `oauth-authz-req+jwt`. The non-normative example is updated accordingly.

### DIFF
--- a/docs/en/remote-flow.rst
+++ b/docs/en/remote-flow.rst
@@ -275,7 +275,7 @@ The JWS header parameters are described below:
   * - **alg**
     - Algorithm used to sign the JWT, according to [:rfc:`7516#section-4.1.1`]. It MUST be one of the supported algorithms in Section *Cryptographic Algorithms* and MUST NOT be set to ``none`` or to a symmetric algorithm (MAC) identifier.
   * - **typ**
-    - Media Type of the JWT, as defined in [:rfc:`7519`] and [:rfc:`9101`]. It MUST be set to the value ``oauth-authz-req+jwt``.
+    - Media Type of the JWT, as defined in [:rfc:`7519`] and [:rfc:`9101`]. It SHOULD be set to the value ``oauth-authz-req+jwt``.
   * - **kid**
     - Key ID of the public key needed to verify the JWS signature, as defined in [:rfc:`7517`]. REQUIRED when ``trust_chain`` is used.
   * - **trust_chain**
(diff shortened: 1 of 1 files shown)


docs/en/remote-flow.rst
```diff
@@ -275,7 +275,7 @@ The JWS header parameters are described below:
   * - **alg**
     - Algorithm used to sign the JWT, according to [:rfc:`7516#section-4.1.1`]. It MUST be one of the supported algorithms in Section *Cryptographic Algorithms* and MUST NOT be set to ``none`` or to a symmetric algorithm (MAC) identifier.
   * - **typ**
-    - Media Type of the JWT, as defined in [:rfc:`7519`].
+    - Media Type of the JWT, as defined in [:rfc:`7519`] and [:rfc:`9101`]. It MUST be set to the value ``oauth-authz-req+jwt``.
```
peppelinux:
```suggestion
    - Media Type of the JWT, as defined in [:rfc:`7519`] and [:rfc:`9101`]. It SHOULD be set to the value ``oauth-authz-req+jwt``.
```

@Zicchio according to JAR and OpenID4VP we don't have the requirement to mandate the value of the typ, that might be JWT or oauth-authz-req+jwt

for this reason, and also to prevent exceptions within the wallet solution evaluating the presentation request objects, we accept the clarification using a more relaxed normative language, also to point the dev attention to this singularity.

SHOULD != MUST, something pretty more than RECOMMEND but definitively the same from the dev perspective.

grausof (approved):

peppelinux (approved):

fmarino-ipzs (approved):

PR #624, opened 2025-04-08
Credential Issuer Solution Refac

### LINKED ISSUE #571: PID/(Q)EAA Issuance Editorials/Discussions
I would like to recommend the following editorial changes to the PID/(Q)EAA Issuance for clarity and consistency. 

- [ ] While in the text, we are always using the term "Wallet Attestation" in Figure 9 we referred to it as "WIA". 
- [ ] Decide to refer to the public key in the Wallet Attestation as cnf or cnf.jwk. 
- [ ] As "Conditional" is not a defined term, we can integrate the logic inline within the text. 
- [ ] While in the SPID/CIE OIDC we use the targeted endpoint as the value of the "aud" claim in IT Wallet, we are using the issuer identifier. For the consistency, we may need to make a decision.

fmarino-ipzs:
This PR:

- creates a new section about Issuer Solution with:
  - Logical Architecture
  - Components description
 
- resolves #571


docs/en/issuer-solution.rst
```diff
@@ -0,0 +1,152 @@
...
+   4. Authenticate to Wallet Instances during issuance to prove its legitimacy.
+   5. Support immediate issuance flow and MAY support deferred issuance for various operational scenarios.
+   6. Implement appropriate error handling and User notifications for all processes.
+   7. Maintain comprehensive audit trails while respecting privacy regulations.
+   8. Issue Digital Credentials that support Selective Disclosure.
+   9. Periodically renew its trust with the Federation.
+   10. Register the Relying Party Component (if required) within the national eIDAS ecosystem (for PID issuance) and IT-Wallet ecosystem (for (Q)EAA issuance).
+   11. For PID issuance, authenticate Users with LoA High using national Digital Identity Providers.
+   12. For (Q)EAA issuance requiring authentication, verify a valid PID from the User's Wallet Instance via `OpenID4VP`_.
+   13. Implement proper procedures for the entire Digital Credential lifecycle as detailed in Section :ref:`Digital Credential Lifecycle`.
```
peppelinux, line 36: for the next release, ideally, all the requirements should be defined with some sort of UID and within a table.

this should be normalized across all the section, and developed/proposed by a single subject.

peppelinux (approved):

docs/en/issuer-solution.rst
peppelinux: moved to a separate issue

grausof (approved):

giadas (approved):

PR #638, opened 2025-04-11
Testplans

peppelinux:
This PR cover the first, very general, configuration of test plans:

- refactors the one about wallet solution
- introduces vci tests
- introduces openid4vp
- introduced proximity tests

many of the test plan should be exploded in multiple one in the next release for providing a more accurate guidance to the implementers.


peppelinux:
Closed, to be reopened with only test plan contents.

Issue #651, opened 2025-04-17
Translate the content in italian


peppelinux:
The current content must be translate in italian using different pull request divided with the following chapters/purposes

1.intro/terms/ux/UI
2. Wallet solution/mobile/embedsed solutions 
3. Pid EAA fata model, issuance and revocation
4. Presentation
5. Trust
6. Outro/contrib

Test plans and technical specifications should not be necessarly be translated

RosaliaGaleano:
There is no need to translate the brand-identity, design and functionality sections. 
These will be covered by me and @steps88

steps88:
There's a draft [PR](https://github.com/italia/eid-wallet-it-docs/pull/661) with a first version of the italian content

peppelinux:
Resolved in the current editor's copy

PR #657, opened 2025-04-18
Design sections translated into Italian

RosaliaGaleano:
## Title
Design sections translated into Italian
## Content
The PR proposes the Italian translation of the design sections (issue #651 )


peppelinux:
this seems invalid, the translated content is also missing

PR #859, opened 2025-10-13
fix: PID discrepancies with PID ARF Rulebook

### LINKED ISSUE #698: PID Rulebook ARF alignment
In [PID Rulebook](https://raw.githubusercontent.com/eu-digital-identity-wallet/eudi-doc-attestation-rulebooks-catalog/refs/heads/main/rulebooks/pid/pid-rulebook.md):

- "**place_of_birth**" contains at least one of the following member: **country**, **region** or **locality**. Currently the claim is a String.
- "**vct**" for all PIDs as a convention must use types in the namespace "**urn:eudi:pid**". Currently vct is defined as a URL.

peppelinux:
This PR resolves https://github.com/italia/eid-wallet-it-docs/issues/698

it also makes iat optional, since the sole entity that must know when a credential was issued is its issuer

wallet instance must keep this information within its internal log

RP doesn't rellay matter


docs/en/credential-data-model.rst
```diff
@@ -113,14 +118,16 @@ The JWT payload contains the following claims. Some of these claims can be discl
       - [NSD]. OPTIONAL. The identifier of the subject of the Digital Credential, the User, MUST be opaque and MUST NOT correspond to any anagraphic data or be derived from the User's anagraphic data via pseudonymization. Additionally, it is required that two different Credentials issued MUST NOT use the same ``sub`` value.
       - `[RFC7519, Section 4.1.2] <https://www.iana.org/go/rfc7519>`_.
     * - **iat**
-      - [SD]. REQUIRED. UNIX Timestamp with the time of JWT issuance, coded as NumericDate as indicated in :rfc:`7519`.
+      - [NSD]. OPTIONAL. UNIX Timestamp with the time of JWT issuance, coded as NumericDate as indicated in :rfc:`7519`.
```
peppelinux, line 121: @m-basili reported that we do not have any clear statement about the presence/absence of `iat` in both the [PID rulebook](https://github.com/eu-digital-identity-wallet/eudi-doc-attestation-rulebooks-catalog/blob/main/rulebooks/pid/pid-rulebook.md) and the IA.

we have issuance_date for mdoc cbor but nothing for sd-jwt. According to this evidence we could keep iat within the payload.

docs/en/registry-catalogue.rst
```diff
@@ -91,6 +91,20 @@ The following table summarizes the main information that MUST be provided by the
...
+- Credential Issuer metadata and Digital Credential Metadata Type MAY still contain display information for interoperability with external ecosystems and for Credentials not registered in the Catalogue; however, when a Credential is present in the Catalogue, Catalogue values MUST take precedence over any display information provided by the Credential Issuer metadata and/or the Digital Credential Metadata Type.
+- For Credentials that are not included in the Catalogue (for example, Credentials not deemed of public interest), Wallet Solutions MAY use the display information from the Credential Issuer metadata and/or the Digital Credential Metadata Type. When both are available, the following precedence applies:
+
+  1. Use the Digital Credential Metadata Type if available.
+  2. Otherwise, use the Credential Issuer metadata.
+
+- Implementations SHOULD cache Catalogue display data and apply language selection using the ``localization`` parameters. If the selected locale is unavailable, fall back to the Catalogue default locale.
+
```
peppelinux, line 107:
```suggestion
```

OOS for this PR

docs/it/registry-catalogue.rst
```diff
@@ -91,6 +91,20 @@ La seguente tabella riassume le principali informazioni che DEVONO essere fornit
...
+- Per gli Attestati Elettronici non inclusi nel Catalogo (ad esempio Attestati Elettronici non di interesse pubblico), le Soluzioni Wallet POSSONO utilizzare le informazioni di visualizzazione dai Metadata del Fornitore di Attestati Elettronici e/o dal Type Metadata dell'Attestato Elettronico. Quando entrambi sono disponibili, si applica la seguente precedenza:
+
+  1. Utilizzare il Type Metadata dell'Attestato Elettronico, se disponibile.
+  2. In caso contrario, utilizzare i Metadata del Fornitore di Attestati Elettronici.
+
+- Le implementazioni DOVREBBERO mettere in cache i dati di visualizzazione del Catalogo e applicare la selezione della lingua utilizzando i parametri di ``localization``. Se la localizzazione selezionata non è disponibile, effettuare il fallback alla localizzazione predefinita del Catalogo.
+
```
peppelinux, line 107:
```suggestion
```

OOS for this PR

docs/en/credential-data-model.rst
peppelinux: @m-basili during the call of 22 ioctober we have decided to have a full compliance to the PID rulebook keeping at the same time the retrocompatibility with the previous jwt metadata

we have decided to add the required parameters even if overlapping with legacy jwt metadata parameters.

We need to add a note that the parameters values MUST use the same value even if it must provided using different data formats (eg: iat uses ISO UTC timestamp, while expiry_date uses another ISO format)

peppelinux: see also the elements below to be considered in this alignment

peppelinux: image

peppelinux: image

PR #951, opened 2025-12-19
Adds official resources folder, updates Brand identity

steps88:
## Review

- [ ] Ensure your files are written following RST specs (*not MD!*)
- [ ] Italian version
- [ ] English version
- [ ] Example files 
- [ ] Ask for review


m-basili (approved):

grausof:
Just a question, but why are all the logos on a white background? Isn't there a negative and positive expected? Also, isn't there even a logo with a transparent background? Thank you

steps88:
> Just a question, but why are all the logos on a white background? Isn't there a negative and positive expected? Also, isn't there even a logo with a transparent background? Thank you

There are multiple variants of logo, but we decided to have only the primary here and redirect to the Brand Manual for everything else

grausof (approved):

PR #945, opened 2025-12-16
Update Digital Credentials Hierarchy and Taxonomy

AntoPaparella:
nota: rivedere tabella per digital credential Hiearchy ed esempio non normativo in Taxonomy. @peppelinux @fmarino-ipzs @grausof le modifiche sono state completate, compresa la "Purpose". A partire da riga 565 esiste una tabella che  non viene renderizzata in preview. Sapete aiutarmi al fine di visualizzarla correttamente? è bloccante per il merge? 

## Review

- [ ] Ensure your files are written following RST specs (*not MD!*)
- [ ] Italian version
- [ ] English version
- [ ] Example files 
- [ ] Ask for review

PR #952, opened 2025-12-19
Taxonomy Review

pcerqua-ipzs:
## Title

Taxonomy Review

## Content

This PR inherits all changes proposed in https://github.com/italia/eid-wallet-it-docs/pull/945

## Review

- [x] Italian version
- [x] English version
- [x] Example files 
- [x] Ask for review


fmarino-ipzs (changes requested):
I see some potential issues in this proposal. The main one is the following: 

The taxonomy defines a four-level hierarchy where Credential Types are the 3rd level (subclass). At the same time, the credential catalog treats these same Credential Types as concrete entries. This creates duplications and confusion. 

The Taxonomy should instead define only abstract semantic categories (Domains, Classes, and Purposes), while Credential Types should exist exclusively in the Catalog and reference the Taxonomy for classification.

I fixed some minor issues, like credential type vs vct vs doctype: 
vct and doctype are format-specific technical identifiers rather than semantic properties of the credential itself. Mixing semantic credential identification (what the credential represents) with technical format identifiers (how it's encoded) creates confusion between the logical concept of a credential type and its technical representations. For example, a Mobile Driving License is semantically the same credential whether encoded as SD-JWT or mDOC-cbor. 

Moreover, I see some potential ambiguity about whether Purposes represent abstract verification intents in the Taxonomy or concrete claim requirement specifications tied to specific credentials in the Catalog. We should decide which approach we want to follow.

Finally, I think that further investigation and analysis are necessary to address these concerns before proceeding.

pcerqua-ipzs:
In order to address the potential issues, we proposed a different technical and logical abstraction of the original proposed hierarchy.

On the *logical side*, we define a three-level hierarchy for Domains, Credential Classes, and Credentials, and a separate mapping between Credential Classes and the respective verification Purposes.

On the *technical side*, the Taxonomy Endpoint defines the hierarchy of Domains, Credential Classes, and Purposes, while Credential Catalog Endpoint contains the mapping between a Credential and the Taxonomy characteristics.

In this proposal, Purposes represent abstract verification intents and those intents are initially defined during AS onboarding, when the intended use for a specific dataset is declared, and then, based on that information, a CI defines the domains, classes and purposes of the specific Credential that will contain, one or more, exposed datasets.

This approach should maintain clarity, expressivity, standardization, and discoverability while avoiding duplication over Registry Infrastructure components.

@fmarino-ipzs @peppelinux  @m-basili

peppelinux:
@AntoPaparella please review ^

AntoPaparella (approved):
grazie per la review

grausof (approved):

peppelinux (approved):

m-basili (approved):

PR #968, opened 2026-01-13
fix: sphinxcontrib-redoc still uses deprecated pkg_resources #960

### LINKED ISSUE #960: sphinxcontrib-redoc still uses deprecated pkg_resources
We may add setuptools<81 to requirements-dev.txt to suppress the deprecation warning. This pins setuptools to a version that doesn't emit the warning, as recommended in the warning message.

This may be temporary workaround until sphinxcontrib-redoc migrates from pkg_resources to importlib.metadata. Once that package is updated, we can remove this pin. 

This issue aims to share this analisys and make us to not forget how to resolve it when the next version of sphinxcontrib-redoc will be available.

pcerqua-ipzs:
## fix: sphinxcontrib-redoc still uses deprecated pkg_resources

fixes #960


fmarino-ipzs (approved):

peppelinux:
The documentation build depends on sphinxcontrib-redoc (v1.6.0), which still imports pkg_resources to read its own version. That API is deprecated and was removed in setuptools 82+.

- With setuptools < 82: builds succeed but emit deprecation warnings.
- With setuptools ≥ 82: the extension fails at import with ModuleNotFoundError: No module named 'pkg_resources'.

No fix has been released. PyPI latest remains 1.6.0 (April 2020). The upstream repository still uses pkg_resources; related issue [#53](https://github.com/sphinx-contrib/redoc/issues/53) is open.

Keep the setuptools pin until upstream replaces pkg_resources (e.g. with importlib.metadata) or we migrate to an alternative OpenAPI/Sphinx renderer.

@pcerqua-ipzs may we evaluate alternatives to OpenAPI/Sphinx renderer?

Issue #980, opened 2026-01-23
Clarification on Public Key distribution mechanism for GetAttributeClaims response verification


gianpaolo-tndigit:
**Context**
According to the API specifications (openapi.json), the `GetAttributeClaims` endpoint returns a `CredentialClaimsResponse` formatted as a signed **JWT** upon success (HTTP 200).
As a Credential Issuer (Consumer), it is mandatory to verify the signature of this token to ensure data integrity and authenticity from the Authentic Source (Provider).

**Current ambiguity**
The current documentation does not explicitly specify the mechanism for retrieving the public key needed to verify the JWT signature.

**Proposed Pattern / Assumption**
Based on the standard PDND interoperability model, we assume the flow should be:

1. The Authentic Source signs the response JWT with its private key.
2. The corresponding public key is stored in the **PDND Provider Keyring** ("Portachiavi Erogatore").
3. The Credential Issuer retrieves the public key via standard PDND API using the `kid` found in the JWT header.

**Request**
Could you please confirm if this is the intended pattern?
If so, we suggest explicitly documenting this in the specifications to rule out other mechanisms (such as OIDC Discovery/.well-known endpoints or out-of-band distribution) and provide clear guidance for implementers.

Thank you.

Reference This issue continues the discussion started here: https://github.com/italia/eid-wallet-it-regioni/discussions/29

m-basili:
yes, like reported in the following [link ](https://italia.github.io/eid-wallet-it-docs/versione-corrente/en/e-service-pdnd.html#e-service-usage:~:text=Step%204%20(e%2DService%20Response))


<img width="925" height="498" alt="Image" src="https://github.com/user-attachments/assets/ea53178e-b019-458a-8b49-fa2dfaf3e6c6" />

gianpaolo-tndigit:
Thanks @m-basili for the confirmation.

I have opened PR #983 to address this.
I thought it was appropriate to explicitly reference the Technical Appendix within the specific endpoint section to ensure the pattern is clear without duplication.
Additionally, I included a specific warning to explicitly rule out alternative key distribution mechanisms (such as public `.well-known` endpoints), to avoid any potential ambiguity for implementers.

peppelinux:
Related PR merged, considered ready closing

PR #983, opened 2026-01-23
docs: clarify signature verification for GetAttributeClaims endpoint

gianpaolo-tndigit:
## Title
Clarification on JWT Signature Verification for GetAttributeClaims

## Content
This PR updates the `authentic-source-endpoint.rst` documentation (both Italian and English versions) to resolve ambiguity regarding the JWT signature verification for the `GetAttributeClaims` endpoint response.

Specifically, it:
- Explicitly states that the Credential Issuer MUST verify the signature of the `CredentialClaimsResponse` JWT.
- Mandates the use of the standard PDND Interoperability model for key retrieval (referencing the Technical Appendix).
- Adds a warning to explicitly forbid alternative key distribution mechanisms (e.g., public `.well-known` endpoints or out-of-band distribution), ensuring trust remains centralized within the PDND infrastructure.

Relates to issue: italia/eid-wallet-it-docs#980

## Review

- [x] Italian version
- [x] English version
- [ ] Example files
- [x] Ask for review

### DIFF
--- a/docs/en/authentic-source-endpoint.rst
+++ b/docs/en/authentic-source-endpoint.rst
@@ -44,7 +44,7 @@ Get Attribute Claims
     - the Authentic Source MUST record the datetime value provided within the ``last_updated`` parameter, which indicates the last time the User's attributes were updated in the Authentic Source's database;
     - the Credential Issuer MUST read the ``last_updated`` value received in the response to be able to check if the User's attributes have changed since the last issuance of a Digital Credential.
 
-The successful response (HTTP 200) returns a ``CredentialClaimsResponse`` object formatted as a **Signed JWT**.
+The successful response (HTTP 200) returns a ``CredentialClaimsResponse`` object formatted as a signed JWT.
 
 Signature Verification and Key Management
 '''''''''''''''''''''''''''''''''''''''''
--- a/docs/en/authentic-source-endpoint.rst
+++ b/docs/en/authentic-source-endpoint.rst
@@ -43,3 +43,16 @@ Get Attribute Claims
     - both MUST save the ``jti`` value in the Agid-JWT-Signature payload of the request to manage Signals related to the deffered issuance of a Digital Credential (see :ref:`signal-hub-endpoint:Signals Processing`);
     - the Authentic Source MUST record the datetime value provided within the ``last_updated`` parameter, which indicates the last time the User's attributes were updated in the Authentic Source's database;
     - the Credential Issuer MUST read the ``last_updated`` value received in the response to be able to check if the User's attributes have changed since the last issuance of a Digital Credential.
+
+The successful response (HTTP 200) returns a ``CredentialClaimsResponse`` object formatted as a **Signed JWT**.
+
+Signature Verification and Key Management
+'''''''''''''''''''''''''''''''''''''''''
+
+As the response token is signed, the Credential Issuer (Consumer) MUST verify the signature to ensure the integrity and authenticity of the data received from the Authentic Source.
+
+The signature verification and key retrieval process MUST strictly follow the standard pattern defined for **PDND e-Services**.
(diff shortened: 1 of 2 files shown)


m-basili (approved):
Although the information is already present in the e-Service PDND section, this PR may improve readability.

docs/en/authentic-source-endpoint.rst
```diff
@@ -43,3 +43,16 @@ Get Attribute Claims
     - both MUST save the ``jti`` value in the Agid-JWT-Signature payload of the request to manage Signals related to the deffered issuance of a Digital Credential (see :ref:`signal-hub-endpoint:Signals Processing`);
     - the Authentic Source MUST record the datetime value provided within the ``last_updated`` parameter, which indicates the last time the User's attributes were updated in the Authentic Source's database;
     - the Credential Issuer MUST read the ``last_updated`` value received in the response to be able to check if the User's attributes have changed since the last issuance of a Digital Credential.
+
+The successful response (HTTP 200) returns a ``CredentialClaimsResponse`` object formatted as a **Signed JWT**.
```
andreamisuraca: @m-basili @peppelinux why is `CredentialClaimsResponse` represented as a JWS? Could the response simply include the payload and be signed according to [this](https://developer.pagopa.it/pdnd-interoperabilita/guides/manuale-operativo-pdnd-interoperabilita/v1.0/tutorial/tutorial-per-lerogatore/come-firmare-una-risposta-per-un-fruitore) specification instead?

peppelinux: @m-basili ^

m-basili: According to Modi (see [section 5.3 of Modi](https://www.agid.gov.it/sites/agid/files/2024-07/Linee_guida_interoperabilit%C3%A0PA_All2_Pattern_sicurezza.pdf)), the INTEGRITY REST 02 pattern, mentioned in your reference to the dev portal section, applies only to the Consumer Request. We also required the integrity property for the Provider Response. That's why we have considered a `CredentialClaimsResponse ` as a Signed JWT. What do you think? @andreamisuraca

andreamisuraca: @m-basili I fully agree with you, ModI requires this for the request. On other occasions, we discussed this and felt it would be better to use the same pattern for both the request and the response. That's why you see `INTEGRITY_REST_02` in the PDND user guide for a response. What do you think? Is there any advantage to using a different pattern for the response?

m-basili: Probably in most practical implementations the use of a JWT tends to be more common and interoperable, especially when dealing with signed responses. However do you know if there are plans to update ModI accordingly, given that it remains the normative reference? @andreamisuraca, @peppelinux

andreamisuraca: @m-basili we will probably propose to introduce this change in the next ModI update for the following reasons:
- We did not identify any specific reason to use different patterns for requests and responses.
- Many administrations are already familiar with this pattern.
- Several administrations already apply it consistently to both requests and responses.
- Many PDND-compatible API gateways, often open source, already support this pattern for both requests and responses. For this reason, we also decided to reference it in the PDND user guide.

Do you think we could adopt the same approach for the AS e-service as well?

m-basili: Yes, we could adopt.
Which content-type is more commonly used by administrations for the responses: `application/json` or `application/jwt`?

grausof (approved):

docs/en/authentic-source-endpoint.rst
andreamisuraca: I would say `application/json` and `application/problem+json` for errors, according to ModI [RAC_GEN_FORMAT_002].

peppelinux:
going to be merged into https://github.com/italia/eid-wallet-it-docs/pull/990

there we'll consolidate all about OAS in it-wallet

PR #995, opened 2026-01-30
docs: use bootstrap italia references

astagi:
This PR removes all the references to `dev-kit-italia` since it's not been used in any example shown in this documentation.

I think that my suggestion about using `dev-kit-italia` has been misunderstood. At the moment `dev-kit-italia`'s not been officially released yet despite you can already install some packages from npm and use them in production (`it-button` and `it-icon` are simple and stable imho), but this is not an operation that will be easy and transparent for the final users. The main risk is that they will be forced to install `dev-kit-italia` packages and set up everything. 

A better way to use `dev-kit-italia` would be creating a "It wallet" web component delivered through both npm and prebuilt bundle that uses `it-button` and `it-icon` components transparently, but I don't know whether it's the goal of this project.

## Review

- [x] Italian version
- [x] English version
- [ ] Example files 
- [x] Ask for review

### DIFF
--- a/docs/en/official-resources.rst
+++ b/docs/en/official-resources.rst
@@ -23,6 +23,13 @@ The following official graphic assets are available for download in the `officia
 - **Trust Mark**: Official graphic element for attesting the participation in the IT-Wallet System - `Download Trust Mark <_static/trustmark-ENG.svg>`_ - Read more at section :ref:`brand-identity:Trust Mark`
 - **Authentication Button**: Official button for authentication via IT-Wallet - `Download Authentication Button <_static/Authentication-button-ENG.svg>`_
 
+HTML Components
+^^^^^^^^^^^^^^^
+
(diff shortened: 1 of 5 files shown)


peppelinux:
we cannot remove the distribution of the official button since it is in the strict scope of the current milestone

peppelinux (changes requested):
the static asset with the authentication button is in the scope of the current milestone

steps88 (approved):

grausof (approved):

grausof:
I agree with @astagi  that a web component for `it-wallet` should be created to be distributed via npm and prebuilt bundles and perhaps placed in a separate repo from this one which instead deals exclusively with technical specifications.

peppelinux (approved):

m-basili (approved):

PR #999, opened 2026-02-04
Fix OAS3-PDND-AS

nardil:
- Removed redundant `CredentialClaimsResponse` 
- Extended authentication to the `/status` endpoint as well
- Fixed response message examples


m-basili (approved):

andreamisuraca (approved):
Auth on `/status` is a choice. This path may be used for PDND probing functionality in the future. Looks good to me.

grausof (approved):

PR #1049, opened 2026-02-24
Update AS OAS3

### LINKED ISSUE #1024: Introduce an AS-managed deferred_object_id for Deferred Flow Retrieval
# Current behavior

In the current deferred flow, when the requested attestation/dataset is not immediately available, the identifier used to retrieve it later is derived from the `jti` value emitted by the Credential Issuer. In practice, clients use this jti-based identifier in subsequent polling/retrieval calls to fetch the final dataset/attestation once it becomes available.

# Problem

This approach is not ideal for two main reasons:

- Unwanted coupling: `jti` is an Issuer-side token identifier. Using it as the retrieval/correlation identifier ties the Authorization Server (AS) deferred flow to an Issuer-internal construct.
- Lack of AS control: In a deferred flow, the AS should be able to define and govern the correlation/retrieval handle independently, without relying on the Issuer’s token semantics.

Overall, using `jti` as a deferred retrieval identifier makes the flow less practical and harder to evolve, and it blurs responsibilities between AS and Issuer.

# Proposed solution

For deferred responses, the AS should explicitly return an opaque retrieval handle generated and managed by the AS. This handle is then used in subsequent calls to retrieve the dataset/attestation. When the dataset becomes available, this identifier maps to (or remains) the anonymous dataset identifier.

Suggested response shape for deferred cases:

```json
{
  "interval": 600,
  "deferred_object_id": "12345-67890-abcdef"
}
```

Where:

- `deferred_object_id` is an opaque identifier created and managed by the AS.
- Clients must use deferred_object_id for later retrieval/polling operations.
- The AS remains the authority for correlation in the deferred flow, removing dependency on the Issuer’s jti.

# Benefits

- Decouples deferred retrieval from Issuer token internals (`jti`)
- Restores clear ownership of deferred correlation to the AS
- Improves maintainability and future extensibility of the deferred flow

m-basili:
This PR resolves #1024.

This PR contains:

- `object_id` MUST be present also in case of deferred issuance since each `object_id` corresponds to a `credential_identifier` that corresponds to a `transaction_id` singularly used in the Deferred Endpoint. So `object_id` replaces `deferred_object_id`. 
- Reorganized claims position between `AttributeClaims `(these claims will be present in the future Credential) and `MetadataClaims `(these  will not be present in the Credential).
- Added `description` and `status_description` in the `metadataClaims`

### DIFF
--- a/docs/en/oas3/OAS3-PDND-AS.yaml
+++ b/docs/en/oas3/OAS3-PDND-AS.yaml
@@ -181,14 +181,14 @@ paths:
                   personal_administrative_number: "12345A123A"
                 attributeClaims:
                   - object_id: "6F9619FF-8B86-D011-B42D-00C04FC964FF"
-                    status: VALID
+                    status: "VALID"
                     last_updated: "2025-01-15T10:30:00Z"
                     institute_name: "Nome Istituto Universitario"
                     programme_type_name: "Laurea Magistrale"
                     degree_course_name: "Computer Science - Informatica"
                     academic_qualification_date: "2025-06-25"
                   - object_id: "7A0720AB-9C97-E122-C53E-11D05FD075GG"
-                    status: VALID
+                    status: "VALID"
                     last_updated: "2025-01-10T08:00:00Z"
                     institute_name: "Nome Istituto Universitario"
                     programme_type_name: "Laurea Triennale"
@@ -411,7 +411,12 @@ components:
                 example: "2025-12-31"
             required: [object_id]
         interval:
-          description: Required if claims parameter is not present. This represents the estimated amount of time (in seconds) required before making the request of the attribute claims again.
+          description: Required if userClaims and AttributeClaim parameters are not present. This represents the estimated amount of time (in seconds) required before making the request of the attribute claims again.
+          type: integer
+          format: int64
+          example: 864000
+        deferred_object_id:
+          description: Required if userClaims and AttributeClaim parameters are not present. An opaque identifier related to the request of User and Attribute claims. 
           type: integer
           format: int64
           example: 864000
@@ -425,7 +430,7 @@ components:
           description: ID ANPR or Tax identification number
         object_id:
           type: string
-          description: Unique identifier of the Credential dataset or `jti` of the Agid-JWT-Signature Credential Issuer deferred flow's request. If this parameter is present only the indicated dataset is returned
+          description: Unique identifier of the Credential dataset or `deferred_object_id` obtained in the previous response. If this parameter is present only the indicated dataset is returned.
     ProblemDetails:
       type: object
       description: RFC7807-compliant problem details object for error responses.
--- a/docs/en/oas3/OAS3-PDND-AS.yaml
+++ b/docs/en/oas3/OAS3-PDND-AS.yaml
@@ -417,9 +417,8 @@ components:
           example: 864000
         deferred_object_id:
           description: Required if userClaims and AttributeClaim parameters are not present. An opaque identifier related to the request of User and Attribute claims. 
-          type: integer
-          format: int64
-          example: 864000
+          type: string
+          example: "CF518451-0E54-4E52-B6C4-F1D952BE0505"
     CredentialClaimsRequest:
       required:
         - unique_id
--- a/docs/en/oas3/OAS3-PDND-AS.yaml
+++ b/docs/en/oas3/OAS3-PDND-AS.yaml
@@ -365,7 +365,7 @@ components:
             type: object
             properties:
               object_id:
-                description: Unique identifier of the Dataset.
+                description: Unique identifier of the Dataset. It MUST NOT contain personal data.
                 type: string
                 example: "6F9619FF-8B86-D011-B42D-00C04FC964FF"
               status:
@@ -398,7 +398,7 @@ components:
             type: object
             properties:
               object_id:
-                description: Unique identifier of the Dataset.
+                description: Unique identifier of the Dataset. It MUST NOT contain personal data.
                 type: string
                 example: "6F9619FF-8B86-D011-B42D-00C04FC964FF"
               issuance_date:
@@ -416,7 +416,7 @@ components:
           format: int64
           example: 864000
         deferred_object_id:
-          description: Required if userClaims and AttributeClaim parameters are not present. A unique opaque identifier related to the request of User and Attribute claims. 
+          description: Required if userClaims and AttributeClaim parameters are not present. A unique opaque identifier related to the request of User and Attribute claims. It MUST NOT contain personal data.
           type: string
           example: "CF518451-0E54-4E52-B6C4-F1D952BE0505"
     CredentialClaimsRequest:
--- a/docs/en/oas3/OAS3-PDND-AS.yaml
+++ b/docs/en/oas3/OAS3-PDND-AS.yaml
@@ -365,12 +365,37 @@ components:
             type: object
             properties:
               object_id:
+                description: Unique identifier of the Dataset. It MUST NOT contain personal data. Required also if additionalProperties are not present. This parameter MUST be used to notify through Signal Hub data updating or availability.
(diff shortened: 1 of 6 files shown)


docs/en/signal-hub-endpoint.rst
```diff
@@ -69,8 +69,8 @@ The Signal Collection e-Service endpoint is used by Authentic Sources to deposit
   * - **objectId**
     - REQUIRED. The subject to which the Signal is bound. If the Signal has ``signalType``:
       
-      - ``CREATE``, then it MUST be set to the ``jti`` value the Credential Issuer used in the Agid-JWT-Signature token of the `get attributes` request to the Authentic Source to obtain the attributes related to a specific Digital Credential (see :ref:`authentic-source-endpoint:Get Attribute Claims`);
-      - ``UPDATE``, then it MUST be set to the Authentic Source's unique database identifier of the Digital Credential's attributes the Signal refers to.
+      - ``CREATE``, then it MUST be set to the ``deferred_object_id`` that the Authentic Source used in the `get attributes` payload response to the the Credential Issuer in order to obtain the attributes related to a specific Digital Credential (see :ref:`authentic-source-endpoint:Get Attribute Claims`);
```
andreamisuraca: @m-basili if I understand correctly, the `deferred_object_id` and the `object_id` refer to the same value, is that right? In the deferred flow, the `object_id` is returned immediately from the AS to the CI, but it is named `deferred_object_id`. Is that correct?
Why did we decide to use a different ID at that time?

docs/en/oas3/OAS3-PDND-AS.yaml
```diff
@@ -411,7 +411,12 @@ components:
                 example: "2025-12-31"
             required: [object_id]
         interval:
-          description: Required if claims parameter is not present. This represents the estimated amount of time (in seconds) required before making the request of the attribute claims again.
+          description: Required if userClaims and AttributeClaim parameters are not present. This represents the estimated amount of time (in seconds) required before making the request of the attribute claims again.
+          type: integer
+          format: int64
+          example: 864000
+        deferred_object_id:
```
andreamisuraca: @m-basili is it correct that this is a integer?

andreamisuraca (commented):
Just a couple of questions @m-basili

docs/en/signal-hub-endpoint.rst
m-basili: No, it is different, because the Authentic Source's response may contain multiple `object_id`: one for each dataset (each one will correspond to a distint Credential) , while `deferred_object_id` is related to the availability of the entire datasets.

docs/en/oas3/OAS3-PDND-AS.yaml
m-basili: in [versione-corrente](https://italia.github.io/eid-wallet-it-docs/versione-corrente/en/[OAS3-PDND-AS](https://italia.github.io/eid-wallet-it-docs/versione-corrente/en/OAS3-PDND-AS.html#tag/credential/operation/attributeClaims).html#tag/credential/operation/attributeClaims) is already defined like an integer. In [openID4VCI](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#name-credential-response) `interval `_"contains a positive number that represents the minimum amount of time in seconds ..."_

andreamisuraca: @m-basili I meant this:
```json
deferred_object_id: 
    description: Required if userClaims and AttributeClaim parameters are not present. An opaque identifier related to the request of User and Attribute claims. 
    type: integer
    format: int64
    example: 864000
```

m-basili: fixed

docs/en/oas3/OAS3-PDND-AS.yaml
andreamisuraca:
```suggestion
                description: Unique identifier of the Dataset. It MUST NOT contain personal data.
```

andreamisuraca:
```suggestion
                description: Unique identifier of the Dataset. It MUST NOT contain personal data.
```

docs/en/oas3/OAS3-PDND-AS.yaml
```diff
@@ -411,10 +411,14 @@ components:
                 example: "2025-12-31"
             required: [object_id]
         interval:
-          description: Required if claims parameter is not present. This represents the estimated amount of time (in seconds) required before making the request of the attribute claims again.
+          description: Required if userClaims and AttributeClaim parameters are not present. This represents the estimated amount of time (in seconds) required before making the request of the attribute claims again.
           type: integer
           format: int64
           example: 864000
+        deferred_object_id:
+          description: Required if userClaims and AttributeClaim parameters are not present. A unique opaque identifier related to the request of User and Attribute claims.
```
andreamisuraca:
```suggestion
          description: Required if userClaims and AttributeClaim parameters are not present. A unique opaque identifier related to the request of User and Attribute claims. It MUST NOT contain personal data.
```

andreamisuraca (commented):
Since the IDs will be transmitted to PDND SignalHub, it is preferable to explicitly state that they must not contain personal data. What do you think? @m-basili

andreamisuraca (approved):

peppelinux (approved):

grausof (approved):

peppelinux (approved):

andreamisuraca (commented):
@m-basili 
- When the object is created, the signal is sent using the same object_id as the newly created object (not a different ID), right?
- The status field is not mandatory in all cases. Why?
- Should issuance_date and expiry_date be required fields?

Just checking to ensure I fully understand, in case any questions come up from the e-service providers.

m-basili:
@andreamisuraca,

- Yes, the object_id(s) is(are) always present even if the specific attribute(s) is(are) not available (deferred flow) and is(are) returned to the Issuer. The object_id(s) is(are) used in signal hub for notifying availability and subsequently updating the data (in case)
- In case of deferred flow, the status is unknown, only object_id(s) and description are returned.
- Some datasets don't have this information. For example, a degree has no expiry (administratively)

docs/en/oas3/OAS3-PDND-AS.yaml
```diff
@@ -423,7 +423,7 @@ components:
                 format: int64
                 example: 864000
             required: [object_id, description]
-      required: [userClaims, attributeClaims, metadataClaims]
```
andreamisuraca, line 426: @m-basili why? When is this not required?

ghost (approved):
e-service dataset clarification:  
- "description" means "dataset description"
- “status” means the actual state of the dataset, mapped according to the status list (possible values: “VALID”, “INVALID”, “SUSPENDED”).
-  “status_description” means a description of the actual state of the dataset as defined by the Authentic Source (AS)
- “last_updated” means the timestamp at which the status/attribute change occurred at dataset (ISO 8601 format, e.g. 2025-01-15T10:30:00Z)

docs/en/oas3/OAS3-PDND-AS.yaml
m-basili: userClaims is not mandatory because credentials may exist that are not strictly associated with a person. For example, a movie ticket.

m-basili: But the PDND templates can render this claim mandatory for specific Credentials.

andreamisuraca: @peppelinux is this ok for you?

andreamisuraca: @m-basili @peppelinux It looks good to me. However, since things are becoming a bit messy, please consider separating the templates in the future, one for the sync flow and another for the deferred flow. If you agree, you can go ahead and resolve the conversation.

PR #1145, opened 2026-06-17
"name" renamed into "label"

m-basili:
This PR renames the following Credential Issuer Metadata:

`openid_credential_issuer.credential_configurations_supported.dc_sd_jwt_pid.credential_metadata.claims.display.name  `  

into

`openid_credential_issuer.credential_configurations_supported.dc_sd_jwt_pid.credential_metadata.claims.display.label`

### DIFF
--- a/docs/en/credential-issuer-metadata.rst
+++ b/docs/en/credential-issuer-metadata.rst
@@ -128,11 +128,12 @@ The *openid_credential_issuer* metadata contains the following claims.
             - **path**: It contains the pointer that specifies the path to a specific claim within the Digital Credential as defined in Appendix C of `OpenID4VCI`_.
             - **mandatory**: Boolean which, when set to `true`, indicates that the Credential Issuer will always include this claim in the issued Credential.
             - **sd**: String indicating whether the claim is selectively disclosable. It MUST be set to `always` if the claim is selectively disclosure or `never` if not.
-            - **display**: Array of objects containing display language properties. Array containing display information about the claim indicated in the ``path``. The array contains an object for each language supported. The parameters that MUST be included are
+            - **display**: Array of objects containing display language properties. Array containing display information about the claim indicated in the ``path``. The array contains an object for each language supported. It contains the following parameters:
 
-                - **label**: String value of a display name for the claim.
-                - **description**: human-readable description for the claim.
-                - **locale**: String value that identifies the language of this object represented as a language tag taken from values defined in *BCP47* :rfc:`5646`. There MUST be only one object for each language identifier.
+                - **name**: REQUIRED. String value of a display name for the claim.
+                - **label**: OPTIONAL. String value of a display name for the claim.
+                - **description**: REQUIRED. human-readable description for the claim.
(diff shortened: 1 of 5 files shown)


gispada:
There seems to be a difference between the Credential Issuer metadata defined by OpenID4VCI and the SD-JWT Type Metadata:
- According to OpenID4VCI the `name` field in `display` is correct [[link](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#claims-description-issuer-metadata)]
- According to SD-JWT VC the `label` field in `display` is correct in the Type Metadata [[link](https://www.ietf.org/archive/id/draft-ietf-oauth-sd-jwt-vc-12.html#section-9.2)]

If we are referring to the Credential Issuer Metadata, I believe OpenID4VCI should be the reference.

grausof:
> There seems to be a difference between the Credential Issuer metadata defined by OpenID4VCI and the SD-JWT Type Metadata:
> 
> * According to OpenID4VCI the `name` field in `display` is correct [[link](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#claims-description-issuer-metadata)]
> * According to SD-JWT VC the `label` field in `display` is correct in the Type Metadata [[link](https://www.ietf.org/archive/id/draft-ietf-oauth-sd-jwt-vc-12.html#section-9.2)]
> 
> If we are referring to the Credential Issuer Metadata, I believe OpenID4VCI should be the reference.

Agree. 
`.well-known/openid-federation` `credential_configurations_supported` follows **OID4VCI**, which uses `name`.
`.well-known/type-metadata` follows **SD-JWT VC**, which uses `label`.

docs/en/credential-issuer-metadata.rst
```diff
@@ -130,7 +130,7 @@ The *openid_credential_issuer* metadata contains the following claims.
             - **sd**: String indicating whether the claim is selectively disclosable. It MUST be set to `always` if the claim is selectively disclosure or `never` if not.
             - **display**: Array of objects containing display language properties. Array containing display information about the claim indicated in the ``path``. The array contains an object for each language supported. The parameters that MUST be included are
 
-                - **name**: String value of a display name for the claim.
```
peppelinux, line 133: ``name`` cannot be removed according to openid4vci 1.0 and 1.1

docs/en/credential-issuer-metadata.rst
```diff
@@ -130,7 +130,7 @@ The *openid_credential_issuer* metadata contains the following claims.
...
+                - **label**: String value of a display name for the claim.
```
peppelinux: we can support the existence of this parameter ``label`` within LTS since it doesn't represent a critical change for interop

peppelinux:
@m-basili please in openid4vci metadata

keep ``name`` as it was previously (as mandatory in the national profile)  and add ``label`` as optional for retrocompatibility for the legacy national implementations

while in vct metadata

keep ``label`` as mandatory  and add ``name`` as optional for retrocompatibility for the legacy national implementations

peppelinux (approved):

grausof (approved):

PR #1167, opened 2026-07-02
"Display" Credential Issuer Metadata

m-basili:
Display properties for claims are mandatory only for those defined in the specific Attestation Rulebook.

### DIFF
--- a/docs/en/credential-issuer-metadata.rst
+++ b/docs/en/credential-issuer-metadata.rst
@@ -124,12 +124,12 @@ The *openid_credential_issuer* metadata contains the following claims.
 
                 - **background_color**: OPTIONAL. String value of a background color of the Digital Credential represented as numerical color values defined in `W3C.CSS-COLOR`_.
 
-          - **claims**: REQUIRED. Array of JSON object each describing how a certain claim related to the Credential MUST be displayed to the User. This Array lists the claims in the order they MUST be displayed by the Wallet. To provide detailed information about the claim, the innermost value MUST contain at least the following parameters. See `OpenID4VCI`_ Section A.3.2.
+          - **claims**: REQUIRED. Array of JSON object each describing how a certain claim related to the Credential MUST be displayed to the User. This Array lists the claims in the order they MUST be displayed by the Wallet. To provide detailed information about the claim, the innermost value contains the following parameters. See `OpenID4VCI`_ Section A.3.2.
 
-            - **path**: It contains the pointer that specifies the path to a specific claim within the Digital Credential as defined in Appendix C of `OpenID4VCI`_.
-            - **mandatory**: Boolean which, when set to `true`, indicates that the Credential Issuer will always include this claim in the issued Credential.
-            - **sd**: String indicating whether the claim is selectively disclosable. It MUST be set to `always` if the claim is selectively disclosure or `never` if not.
-            - **display**: Array of objects containing display language properties. Array containing display information about the claim indicated in the ``path``. The array contains an object for each language supported. It contains the following parameters:
+            - **path**: REQUIRED. It contains the pointer that specifies the path to a specific claim within the Digital Credential as defined in Appendix C of `OpenID4VCI`_.
+            - **mandatory**: REQUIRED. Boolean which, when set to `true`, indicates that the Credential Issuer will always include this claim in the issued Credential.
+            - **sd**: REQUIRED.String indicating whether the claim is selectively disclosable. It MUST be set to `always` if the claim is selectively disclosure or `never` if not.
+            - **display**: CONDITIONAL. REQUIRED only for the claims that are defined in the specific Attestation Rulebook. Array of objects containing display language properties. Array containing display information about the claim indicated in the ``path``. The array contains an object for each language supported. It contains the following parameters:
 
                 - **name**: REQUIRED. String value of a display name for the claim.
                 - **label**: OPTIONAL. String value of a display name for the claim.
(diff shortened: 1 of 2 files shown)


docs/en/credential-issuer-metadata.rst
```diff
@@ -124,12 +124,12 @@ The *openid_credential_issuer* metadata contains the following claims.
...
-            - **display**: Array of objects containing display language properties. Array containing display information about the claim indicated in the ``path``. The array contains an object for each language supported. It contains the following parameters:
+            - **path**: REQUIRED. It contains the pointer that specifies the path to a specific claim within the Digital Credential as defined in Appendix C of `OpenID4VCI`_.
+            - **mandatory**: REQUIRED. Boolean which, when set to `true`, indicates that the Credential Issuer will always include this claim in the issued Credential.
+            - **sd**: REQUIRED.String indicating whether the claim is selectively disclosable. It MUST be set to `always` if the claim is selectively disclosure or `never` if not.
+            - **display**: CONDITIONAL. REQUIRED only for the claims that are defined in the specific Attestation Rulebook. Array of objects containing display language properties. Array containing display information about the claim indicated in the ``path``. The array contains an object for each language supported. It contains the following parameters:
```
m-basili:
```suggestion
            - **display**: CONDITIONAL. REQUIRED only for the claims that are shown to the User. Array of objects containing display language properties. Array containing display information about the claim indicated in the ``path``. The array contains an object for each language supported. It contains the following parameters:
```

docs/it/credential-issuer-metadata.rst
```diff
@@ -124,12 +124,12 @@ I Metadata *openid_credential_issuer* contiene i seguenti *claims*.
...
+            - **path**: OBBLIGATORIO. Contiene il puntatore che specifica il percorso all'attributo specifico all'interno dell'Attestato Elettronico come definito nell'Appendice C di `OpenID4VCI`_.
+            - **mandatory**: OBBLIGATORIO. Valore booleano che, se impostato su `true`, indica che il Credential Issuer includerà sempre questo attributo nelle Credenziali che emette.
+            - **sd**: OBBLIGATORIO. Stringa che indica se il claim è divulgabile selettivamente. DEVE essere impostato su `always` se il claim è divulgabile selettivamente o `never` se non lo è.
+            - **display**: CONDIZIONALE. OBBLIGATORIO solo per i *claims* definiti nello specifico Attestation Rulebook. Array di oggetti contenenti le proprietà di visualizzazione. Array contenente informazioni di visualizzazione relative al claim indicato nel ``path``. L'array contiene un oggetto per ogni lingua supportata. Contiene i seguenti parametri:
```
m-basili:
```suggestion
            - **display**: CONDIZIONALE. OBBLIGATORIO solo per i *claims* che vengono mostrati all'Utente. Array di oggetti contenenti le proprietà di visualizzazione. Array contenente informazioni di visualizzazione relative al claim indicato nel ``path``. L'array contiene un oggetto per ogni lingua supportata. Contiene i seguenti parametri:
```

docs/en/credential-issuer-metadata.rst
```diff
@@ -124,12 +124,12 @@ The *openid_credential_issuer* metadata contains the following claims.
 
                 - **background_color**: OPTIONAL. String value of a background color of the Digital Credential represented as numerical color values defined in `W3C.CSS-COLOR`_.
 
-          - **claims**: REQUIRED. Array of JSON object each describing how a certain claim related to the Credential MUST be displayed to the User. This Array lists the claims in the order they MUST be displayed by the Wallet. To provide detailed information about the claim, the innermost value MUST contain at least the following parameters. See `OpenID4VCI`_ Section A.3.2.
+          - **claims**: REQUIRED. Array of JSON object each describing how a certain claim related to the Credential MUST be displayed to the User. This Array lists the claims in the order they MUST be displayed by the Wallet. To provide detailed information about the claim, the innermost value contains the following parameters. See `OpenID4VCI`_ Section A.3.2.
```
manuraf: it would be better to reference Section B.2 instead, since Section A.3.2 is specific to the SD-JWT VC format


```suggestion
          - **claims**: REQUIRED. Array of JSON object each describing how a certain claim related to the Credential MUST be displayed to the User. This Array lists the claims in the order they MUST be displayed by the Wallet. To provide detailed information about the claim, the innermost value contains the following parameters. See `OpenID4VCI`_ Section B.2.
```

docs/en/credential-issuer-metadata.rst
```diff
@@ -124,12 +124,12 @@ The *openid_credential_issuer* metadata contains the following claims.
...
 
-            - **path**: It contains the pointer that specifies the path to a specific claim within the Digital Credential as defined in Appendix C of `OpenID4VCI`_.
-            - **mandatory**: Boolean which, when set to `true`, indicates that the Credential Issuer will always include this claim in the issued Credential.
-            - **sd**: String indicating whether the claim is selectively disclosable. It MUST be set to `always` if the claim is selectively disclosure or `never` if not.
-            - **display**: Array of objects containing display language properties. Array containing display information about the claim indicated in the ``path``. The array contains an object for each language supported. It contains the following parameters:
+            - **path**: REQUIRED. It contains the pointer that specifies the path to a specific claim within the Digital Credential as defined in Appendix C of `OpenID4VCI`_.
+            - **mandatory**: REQUIRED. Boolean which, when set to `true`, indicates that the Credential Issuer will always include this claim in the issued Credential.
```
manuraf, line 130: Section B.2 only marks the claim path as required. We could mark `mandatory` as optional .. in some examples `mandatory` is not included

m-basili: In the technical specification, we have only the following example: 

Image.

The European implementation uses the Metadata Type to validate a credential in SD-JWT during a credential Presentation. So I think that it is important to always have this claim.

manuraf: I noticed that in mdoc isn't always included, for ex. [here](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#name-credential-issuer-metadata-5) .. However, mark as required in it-wallet  would not be an issue for interoperability, so it's ok for me

manuraf (approved):

peppelinux (approved):

PR #1201, opened 2026-08-03
introduction of the IT-Wallet ID

### LINKED ISSUE #1200: Introduction of NPID (IT-Wallet ID) to replace the PID
We should introduce the NPID Credential as an EAA (IT-Wallet ID) to replace the PID.

m-basili:
This PR resolves #1200.


docs/en/defined-terms.rst
RosaliaGaleano: @peppelinux is it a typo? there are two Person Identification Data definition. Moreover, here we are saying that the PID is an electronic attestation..

peppelinux: good catch, thank you, resolved here: ffa13706d88915f197ab1c965e9a6a6d5cc5eb62

peppelinux (approved):

fmarino-ipzs:
This PR and the related ones have been merged, even though we explicitly expressed [concerns ](https://github.com/italia/eid-wallet-it-docs/pull/1202#pullrequestreview-4864531580) about the inclusion of identity matching amongst the PR’s scopes.

- It has been introduced in a note within PID EUDI Section saying that it is NOT EUDI specific, while the mechanisms defined are EUDI specific.
- An explicit reference to identity matching has been introduced into the function for erasing data from the RP, while it is currently not required as we have IT-Wallet ID carrying tax_id_code in its data model. The current LTS version of the IT-Wallet Specification does not support the issuance or usage of the EUDI PID. only the IT-Wallet ID is supported.

PR #1212, opened 2026-09-14
feat: composition of issuer_state in detail

### LINKED ISSUE #1210: Further clarification on issuer_state encryption and URN encoding
In the Credential Offer (credential-issuance-low-level.rst) the spec says:

1. `issuer_state` is an **opaque string** (OpenID4VCI).
2. It **MUST be a URN** `urn:it-wallet:credential-offer:{authenticSourceId}:{datasetId}[:{objectId}]`.
3. That URN **MUST be encrypted** with *“the PDND public key related to the GetAttributeClaims e-service Consumer”*.
4. Non-normative examples put a compact token (`eyJhbGciOiJSU0Et...`) in `issuer_state`.

`authenticSourceId` MUST equal the Authentic Source `entity_id` in the Authentic Source Registry (`https://{organization_domain}[/{optional_path}]`).

The Wallet only echoes the value in the Authorization Request / PAR. The Credential Issuer decrypts it and uses the three identifiers to call `GetAttributeClaims`.

## Problem

The encryption sentence and the URN template may be not enough clear for the implementers.

### 1. JOSE profile missing

Not specified:

- serialization (JWE Compact shoyuld be specified)
- `alg` / `enc` (examples start with `RSA-…` while `algorithms.rst` for it-wallet lists `RSA-OAEP-256`, `ECDH-ES`, `A128GCM` / `A256GCM` ... We should clarify to use PDND for this')
- JWE payload: raw URN string, or an encrypted JWT with claims?
- required JOSE headers (`kid`, `typ`, `cty`)

Today the text says both “MUST be a URN” and “MUST be encrypted”: the **parameter value** is the ciphertext, the URN is the **plaintext**. That layering should be written explicitly.

### 2. Which Consumer key / `kid`

“PDND public key of the GetAttributeClaims Consumer” = CI public key (Provider = Authentic Source, Consumer = Credential Issuer).

Not specified:

- which PDND client (e-service client of the CI as `GetAttributeClaims` fruitor)
- how the offer issuer (CI or AS) obtains the `kid`
- whether `GET /keys/{kid}` on the PDND Interoperability API is the discovery path (as elsewhere in `e-service-pdnd.rst`)

Without a `kid` (or an equivalent discovery rule) the AS cannot encrypt.

### 3. PDND keys are signing keys, not encryption keys

Participant keys in `e-service-pdnd.rst` are `kty: EC`, `crv: P-256`, `key_ops: ["sign"]`. Those cannot do `RSA-OAEP`. `ECDH-ES` would need `use: enc` / `key_ops` for key agreement.

The example `alg` (`RSA-…`) does not match the keys PDND actually exposes. Either:

- the CI MUST register a dedicated encryption JWK on that PDND client, or
- the key is not a PDND participant key (e.g. an `enc` JWK in the CI Entity Configuration),

and the examples must match the chosen profile.

### 4. URN is not parseable if `authenticSourceId` is an `https://` `entity_id`

`entity_id` contains `:`. Splitting

`urn:it-wallet:credential-offer:https://as.example.it/path:education_degree:6F9619FF-…`

on `:` is ambiguous. Percent-encoding of `authenticSourceId` (and of any other segment that may contain `:`) is not specified; an empty trailing segment is forbidden, but encoding is not defined.

A JWT payload with discrete claims (`authentic_source_id`, `dataset_id`, `object_id`) would avoid this; if the URN is kept, encoding MUST be normative.

## Proposed direction (to agree with @fmarino-ipzs @m-basili @andreamisuraca )

To be confirmed, not assumed:

1. State the two layers: plaintext = URN (or JWT claims); `issuer_state` parameter = JWE Compact of that plaintext.
2. Fix `alg`/`enc`/`typ`/`kid` against `algorithms.rst` and the key actually published for the CI.
3. Define key discovery: PDND `GET /keys/{kid}` **or** CI `jwks` with `use: enc`, and how the AS learns the `kid`.
4. Resolve key type: dedicated PDND `enc` key vs federation `enc` JWK; drop RSA from examples if the key is EC.
5. If the URN is kept: percent-encode `authenticSourceId` (RFC 3986) so `:` / `/` do not break parsing. If a JWT payload is chosen, the URN template can remain informative only.
6. Replace truncated examples with one complete (fake) JWE: header, plaintext, and the value in the Credential Offer.
7. Align IT/EN and the CI test plan (today `issuer_state` is OPTIONAL in the flow, REQUIRED in the test plan).

## Files

- `docs/it/credential-issuance-low-level.rst`
- `docs/en/credential-issuance-low-level.rst`
- examples under Credential Offer
- possibly `docs/*/e-service-pdnd.rst`, `docs/*/algorithms.rst`, `docs/*/test-plans-credential-issuer.rst`

## References

- OpenID4VCI §4.1.1 `issuer_state`
- Authentic Source Registry `entity_id` / `dataset_id`
- `GetAttributeClaims` (`OAS3-PDND-AS.yaml`, path `{datasetId}`, optional `object_id`)
- PDND key retrieval (`GET /keys/{kid}`)

peppelinux:
This PR resolves https://github.com/italia/eid-wallet-it-docs/issues/1210, it:

- Clarifies issuer_state as two layers: percent-encoded URN plaintext and JWE Compact (ECDH-ES / A256GCM) encrypted to a dedicated PDND enc key of the GetAttributeClaims Consumer.
- Defines the profile once in the Credential Offer flow, with PDND key registration/discovery in the PDND section; PAR, endpoint, and CI tests only reference that definition.
- Aligns IT/EN, replace truncated RSA examples with a complete JWE, and fix CI_013 so issuer_state / authorization_server are required only when the spec says so.


docs/en/credential-issuance-endpoint.rst
```diff
@@ -132,8 +132,8 @@ The ``request`` JWT payload contained in the HTTP POST message is given with the
       - Unique identifier of the JWT that, together with the value contained in the ``iss`` claim, prevents the reuse of the JWT (replay attack). Since the `jti` value alone is not collision resistant, it MUST be identified uniquely together with its issuer.
       - [:rfc:`7519`].
     * - **issuer_state**
-      - It MUST be present only in case of issuer initiated flow. It MUST contain the same value contained in the Credential Offer.
-      - [:rfc:`7519`].
+      - It MUST be present in the Request Object when the Credential Offer contains ``issuer_state``. It MUST contain the same value contained in the Credential Offer. The format of that value is defined in :ref:`credential-issuance-low-level:issuer_state Parameter`.
```
m-basili:
```suggestion
      - It MUST be present only in issuer initiated flows and when the Credential Offer contains ``issuer_state``. It MUST contain the same value contained in the Credential Offer. The format of that value is defined in :ref:`credential-issuance-low-level:issuer_state Parameter`.
```

docs/en/credential-issuance-low-level.rst
```diff
@@ -690,46 +690,138 @@ The Credential Offer object is a JSON object containing the parameters defined i
...
 
 .. note::
   When using ``credential_offer_uri`` (by reference), the Credential Issuer or third party SHOULD use a unique URI for each Credential Offer or otherwise prevent caching of the URI, as recommended in Section 4.1.3 of [`OpenID4VCI`_].
 
+issuer_state Parameter
+^^^^^^^^^^^^^^^^^^^^^^
+
+The ``issuer_state`` parameter is an opaque string as defined in Section 4.1.1 of [`OpenID4VCI`_].
+
+It is REQUIRED when the Credential Offer is bound to a Credential Dataset of an Authentic Source. Otherwise it is OPTIONAL. When present, the Wallet Instance MUST include the same value in the subsequent Authorization Request, as defined in :ref:`credential-issuance-endpoint:Pushed Authorization Request Endpoint`. The Wallet Instance MUST treat ``issuer_state`` as opaque and MUST NOT decrypt or parse it.
```
m-basili: This information is already indicated on line 693.

peppelinux: thx

going to remove it

peppelinux: done in commit 4e33a87eeba03e6431444e090d5e8d3c09c22313

docs/en/credential-issuance-low-level.rst
```diff
@@ -690,46 +690,136 @@ The Credential Offer object is a JSON object containing the parameters defined i
...
+  * - **epk**
+    - REQUIRED. Ephemeral public key used for ``ECDH-ES``.
+    - :rfc:`7518#section-4.6.1.1`.
+
+The ``cty`` header parameter MUST NOT be present.
+
+Encryption Key
+""""""""""""""
+
+The JWE MUST be encrypted to the PDND encryption public key of the Consumer of the ``GetAttributeClaims`` e-service, that is the Credential Issuer identified by ``credential_issuer``.
```
andreamisuraca, line 781: It isn't clear to me how the Issuer is expected to communicate the kid to the Authentic Source. Note that an e-service consumer can configure multiple clients on PDND, and multiple keys can be loaded on each client. Likewise, e-service producers can load multiple keys on PDND through a dedicated feature called "Producer keychains". In both cases, however, these keys are intended for signing, not for encryption.

This is consistent with the [statement](https://italia.github.io/eid-wallet-it-docs/prs/pr1212/en/e-service-pdnd.html#getattributeclaims-consumer-encryption-keys): "Participant keys registered for PDND Voucher issuance and request integrity are signing keys, as in the non-normative example above. Those keys MUST NOT be used to encrypt issuer_state."
How should this be interpreted @peppelinux?

peppelinux: > however, these keys are intended for signing, not for encryption

this is the argument about evaluating the use of encription key for the specific purpose of encryption.

Issue #1057, opened 2026-03-16
Split Authentic Source OpenAPI specification to distinguish synchronous and deferred flows


andreamisuraca:
Several administrations have contacted us reporting difficulties in distinguishing which fields in the OpenAPI specification apply to the synchronous flow and which apply to the deferred flow.

To improve clarity and usability, it would be preferable to provide two separate OpenAPI specifications, one for the synchronous flow and one for the deferred flow. This approach would allow each administration to adopt the specification relevant to their implementation and would also enable fields to be explicitly defined as required or optional depending on the specific type of e-service.

m-basili:
@andreamisuraca 
Currently, the PDND Authentic Source e-service is unique: if the data is available, it is returned immediately; otherwise, the `interval` claim is returned, and the e-service will later provide the data.

If we split the OpenAPI into two, the e-service would necessarily have to be split into two distinct e-services or endpoints. One returns `interval` and takes into account the issuance request, while the second provides the data.

What do you think about this?

andreamisuraca:
@m-basili I agree with you. We should find a better way to explain to the AS when that field is required. We currently get a lot of requests asking us to clarify this.

PR #1214, opened 2026-09-14
fix: OAS3 and DPoP and security list disambiguation

### LINKED ISSUE #1149: [OAS3-PDND-AS] Security scheme does not reflect IT-Wallet mandatory requirements for Authorization and DPoP/TPoP headers
## Problem description

The current top-level `security` definition in `OAS3-PDND-AS.yaml` (lines 17–19) is:

```yaml
security:
  - BearerAuth: []
  - DPoPAuth: []
```

In OpenAPI 3.x, a top-level `security` list is evaluated as a **logical OR**: a client may authenticate using *either* `BearerAuth` *or* `DPoPAuth`. This does not match the IT-Wallet specification for PDND e-Service calls, which requires:

1. The `Authorization` header with a PDND Voucher is **always mandatory**.
2. The `DPoP` header (DPoP proof JWT, RFC 9449) is **mandatory in the POP_DPoP implementation** and absent in POP_TPoP.
3. In the POP_TPoP implementation, the `Agid-JWT-TrackingEvidence` header acts as proof of possession and is **mandatory**.

The current scheme incorrectly implies that the caller may choose between a Bearer voucher and a DPoP header as alternative authentication mechanisms, which is semantically wrong and potentially misleading for implementors.

Additionally, OpenAPI 3.x has no native mechanism to express conditional header requirements (i.e., "if scheme X is used, then header Y is required"). This is a fundamental limitation of the specification format and must be acknowledged in any proposed fix.

---

## Note on `Agid-JWT-TrackingEvidence` and AUDIT_REST_02

The IT-Wallet specification cites AUDIT_REST_02 as **optional**: the Provider *may* require additional tracked data from the Consumer's domain; if so, this pattern MUST be used. Since Authentic Source e-services do not require AUDIT_REST_02, `Agid-JWT-TrackingEvidence` has no role as audit data in this context.

In the POP_TPoP implementation, `Agid-JWT-TrackingEvidence` is repurposed as a proof of possession mechanism (as an alternative to DPoP when the PDND infrastructure does not support RFC 9449). In the POP_DPoP implementation, proof of possession is already provided by the `DPoP` header, and since AUDIT_REST_02 is not required for Authentic Source e-services, `Agid-JWT-TrackingEvidence` serves no purpose and should be removed entirely from the POP_DPoP variant.

---

### Option A — Single OAS file: mandatory `Authorization`, DPoP documented as an optional header parameter

Fix the `security` block to only reference `BearerAuth` (the voucher is always present) and enrich the `description` of the `BearerAuth` scheme to clarify that the actual `Authorization` scheme value is `DPoP` (not `Bearer`) when using POP_DPoP.

The `DPoP` header parameter on the operation already exists with `required: false`. Its `description` should be updated to explicitly state that it is **mandatory when POP_DPoP is in use** and absent in POP_TPoP.

```yaml
# Top-level security: voucher is always required
security:
  - BearerAuth: []

components:
  securitySchemes:
    BearerAuth:
      type: http
      scheme: bearer
      bearerFormat: JWT
      description: >-
        PDND Voucher. In accordance with the IT-Wallet specification,
        when using the POP_DPoP implementation (RFC 9449), the actual
        scheme value in the `Authorization` header MUST be `DPoP` (not
        `Bearer`), and the `DPoP` header containing the proof JWT MUST
        be present. When using the POP_TPoP implementation, the scheme
        is `Bearer` and the `Agid-JWT-TrackingEvidence` header acts as
        proof of possession.
```

```yaml
# On the operation: DPoP header parameter updated
- name: DPoP
  in: header
  required: false
  description: >-
    DPoP proof JWT (RFC 9449). REQUIRED when the PDND Voucher has been
    issued with DPoP binding (POP_DPoP implementation): in this case the
    scheme in the `Authorization` header is `DPoP`. NOT present in the
    POP_TPoP implementation, where proof of possession is provided by
    the `Agid-JWT-TrackingEvidence` header.
  schema:
    type: string
    format: JWT
```

**Pros:** single file, minimal change, backward compatible with PDND tooling.  
**Cons:** the distinction between POP_DPoP and POP_TPoP remains prose-only; machine-readable validators cannot enforce the conditionality.

---

### Option B — Two separate OAS files: `OAS3-PDND-AS-DPoP.yaml` and `OAS3-PDND-AS-TPoP.yaml`

Provide one dedicated file per implementation variant. This is the only approach that produces fully unambiguous, machine-readable specifications.

The `security` scheme in both files models only the `Authorization` header, which is all OAS3 supports for HTTP authentication schemes. The behavioral difference between the two variants is captured entirely at the operation parameter level: the `DPoP` header is `required: true` in the DPoP variant and absent in the TPoP variant; `Agid-JWT-TrackingEvidence` is `required: true` in the TPoP variant and absent in the DPoP variant.

**`OAS3-PDND-AS-DPoP.yaml`** — POP_DPoP variant (RFC 9449):

```yaml
security:
  - BearerAuth: []

components:
  securitySchemes:
    BearerAuth:
      type: http
      scheme: bearer       # OAS3 has no native "dpop" scheme value
      bearerFormat: JWT
      description: >-
        PDND Voucher with DPoP binding (RFC 9449). The effective scheme
        value in the `Authorization` header MUST be `DPoP`. The `DPoP`
        header containing the DPoP proof JWT is REQUIRED.
```

Operation-level headers for this variant:

```yaml
- name: DPoP
  in: header
  required: true
  description: >-
    DPoP proof JWT (RFC 9449). Mandatory in the POP_DPoP implementation.
  schema:
    type: string
    format: JWT

# Agid-JWT-TrackingEvidence: omitted.
# AUDIT_REST_02 is optional per IT-Wallet specification and is not
# required for Authentic Source e-services. In POP_DPoP, proof of
# possession is already provided by the DPoP header; therefore
# Agid-JWT-TrackingEvidence serves no purpose in this variant.
```

---

**`OAS3-PDND-AS-TPoP.yaml`** — POP_TPoP variant:

```yaml
security:
  - BearerAuth: []

components:
  securitySchemes:
    BearerAuth:
      type: http
      scheme: bearer
      bearerFormat: JWT
      description: >-
        PDND Voucher (Bearer). Proof of possession is provided via the
        `Agid-JWT-TrackingEvidence` header (POP_TPoP implementation).
```

Operation-level headers for this variant:

```yaml
# DPoP: omitted. Not applicable in POP_TPoP.

- name: Agid-JWT-TrackingEvidence
  in: header
  required: true
  description: >-
    JWT acting as proof of possession of the PDND Voucher (POP_TPoP
    implementation), used when the PDND infrastructure does not support
    RFC 9449. Also carries tracked data as per AUDIT_REST_02 if required
    by the Provider; this is not the case for Authentic Source e-services,
    where AUDIT_REST_02 is not mandated.
  schema:
    type: string
    format: JWT
```

**Pros:** each file is self-contained and unambiguous; validators can correctly enforce header requirements for each variant; consistent with the PDND catalog model where each e-service publishes its own OAS.  
**Cons:** two files to maintain; implementors must choose the correct one for their PDND infrastructure configuration.

---

## Notes on OpenAPI 3.x limitations

OpenAPI 3.x `securitySchemes` do not support conditional logic: it is not possible to express "if security scheme X is selected, then header Y is required" in a machine-readable way. Furthermore, a `securityScheme` of type `http/bearer` models only the `Authorization` header — additional headers such as `DPoP` cannot be made mandatory through the security scheme mechanism and must be modeled as operation-level parameters. The OR semantics of a top-level `security` list cannot capture the IT-Wallet requirement that the PDND Voucher in `Authorization` is unconditionally mandatory. These limitations apply to any single-file approach and must be compensated with clear prose documentation.

---

## Recommendation

**Option A** as an immediate minimal fix to remove the misleading OR semantics and align the existing file with the specification.  
**Option B** as the definitive solution, particularly if the intent is for Authentic Sources to publish the correct OAS variant for their chosen implementation in the PDND catalog.

peppelinux:
This PR fixes #1149.

The OpenAPI 3.x `security` list is a logical OR. The previous definition (`BearerAuth` **or** `DPoPAuth`) allowed a DPoP header without a PDND Voucher, or a Bearer Voucher without proof of possession. That contradicts `e-service-pdnd.rst`: `Authorization` is always mandatory, and PoP is either `POP_DPoP` (RFC 9449) or `POP_TPoP` (`Agid-JWT-TrackingEvidence`).

This PR uses Option A from the issue, keep a single OAS file, model only the always-required Voucher in `security`, and document DPoP vs TPoP (and optional `AUDIT_REST_02`) on the operation header parameters. 

This PR does not split into `OAS3-PDND-AS-DPoP.yaml` / `OAS3-PDND-AS-TPoP.yaml` (those are variants of the
same e-service, not two APIs; see also #1057). No three-way AND/OR security block: OAS 3.0 cannot natively express `Authorization: DPoP`, and combining `DPoP` as both `securityScheme` and parameter duplicates the header in codegen.
The same OR bug existed on the Wallet Provider template; it is fixed there too.

## Changelog

- `security` is now only `BearerAuth` (AS top-level; WP `PATCH /wallet-instances`).
- Removed `DPoPAuth` from `securitySchemes`.
- Clarified `BearerAuth`: voucher always required; actual scheme is `DPoP` or `Bearer`.
- Clarified `DPoP` and `Agid-JWT-TrackingEvidence` (`required: false`, prose MUST/MUST NOT).
- AS `info.description`: `REST_JWS_2021_POP` instead of “DPoP as alternative to Bearer”; `AUDIT_REST_02` remains optional.

@nardil please review accordingly


andreamisuraca (commented):
I'm not fully convinced by this. Modelling the DPoP case as `scheme: bearer` doesn't seem semantically correct: the `Authorization` scheme is `DPoP`, not `Bearer`. In practice this also has a concrete cost, Swagger UI and most generators will emit `Authorization: Bearer <voucher>` for the POP_DPoP profile, which a conformant Provider has to reject.

What do you think @peppelinux?

peppelinux:
> A side clarification on this fix: the OpenAPI included here remains a baseline for whoever implements the service. When publishing on PDND, the provider will need to specialize it — both in the attribute_claims returned and in the ModI patterns actually adopted (DPoP vs TPoP) — so that it clearly shows the consumer (eg. IPZS) how the service must be queried. If this reading is correct, the proposed change resolves the issue.

May I ask you to propose where this note should be included? I'd like to include this remark.

peppelinux:
> I'm not fully convinced by this. Modelling the DPoP case as `scheme: bearer` doesn't seem semantically correct: the `Authorization` scheme is `DPoP`, not `Bearer`. In practice this also has a concrete cost, Swagger UI and most generators will emit `Authorization: Bearer <voucher>` for the POP_DPoP profile, which a conformant Provider has to reject.
> 
> What do you think @peppelinux?

yeah, bearer != dpop

I want to get back on this to update the proposal, thx @andreamisuraca

peppelinux:
@nardil I have to do a step back giving the review and the comments from @andreamisuraca 

Before (at https://github.com/italia/eid-wallet-it-docs/issues/1149): security was a logical OR of BearerAuth and DPoPAuth. That read as “Bearer voucher or DPoP header”, so a client could omit the PDND Voucher or skip proof of possession. It also treated DPoP as an alternative to Bearer, which contradicts the ST (Authorization is always required; PoP is POP_DPoP or POP_TPoP).

First fix (Option A): one scheme: bearer scheme and PoP headers as optional parameters. That removed the false OR, but it was still wrong for POP_DPoP: Swagger UI and most generators emit Authorization: Bearer <voucher>, which a conformant Provider must reject (token_type / scheme is DPoP).

Now (current commits at the date of this message): two complete profiles, OR between them, AND inside each:

- Authorization: DPoP and header DPoP
- Authorization: Bearer and Agid-JWT-TrackingEvidence
- PoP headers are security schemes, not duplicate operation parameters. Integrity headers (Agid-JWT-Signature, Digest) are unchanged.

## Why 

We want to match RFC 9449 and e-service-pdnd.rst, by keeping a single OAS file.
We let generators that honour the scheme with DPoP using the right prefix.
We let validators require Voucher + PoP together. 

Implementation consideration: Swagger UI still does not auto-apply non-Bearer HTTP schemes, and it cannot mint a DPoP proof.

docs/en/oas3/OAS3-PDND-AS.yaml
```diff
@@ -124,19 +111,6 @@ paths:
           schema:
             type: string
             example: SHA-256=72e18bdddf13c911b4dd562ee21979a5c9f235c3a01bd1426e857d8c1a282f41
-        - name: Agid-JWT-TrackingEvidence
```
andreamisuraca, line 127: @peppelinux could this header also be used with DPoP? For cases where the authentic source needs additional details from the issuer, like caller LoA, IP address, and so on.

peppelinux: Yes. 

Agid-JWT-TrackingEvidence can also be sent with DPoP when the Authentic Source requires AUDIT_REST_02 (LoA, userID, IP, or any other tracked data agreed with the Provider). 

The header is mandatory for AUDIT_REST_02 or POP_TPoP, and the DPoP proof stays the PoP. 

If an AS actually requires audit data on DPoP, it adds TrackingEvidenceAuth to that pair when specializing the OAS on PDND.

andreamisuraca (approved):
LGTM. Maybe consider adding Agid-Tracking-Evidence as an optional header?

peppelinux:
> LGTM. Maybe consider adding Agid-Tracking-Evidence as an optional header?

done in 83224da67ea4f341ba9b3e79c9345c04fa08e4ed

PR #1220, opened 2026-09-17
federation: wallet_solution -> openid_wallet_provider

### LINKED ISSUE #1219: Wallet Provider federation metadata type is openid_wallet_provider
Wallet Provider federation metadata type is openid_wallet_provider , not wallet_solution

peppelinux:
This PR aligns the specification at openid federation wallet architecture draft about the wallet provider metadata type.

Since this change has impacts on the LTS employments, the deploy should replicate the legacy and the new parameters to not introduce breaking changes.


peppelinux:
@giadas @manuraf this is a breakign change for standard alignments

a release changelog is going to detail this breaking change and the approach to support previous implementations, duplicating the old and the new parameter name, using the same payload.

